PR #83, opened 2016-11-04
registry: Use `is_none` instead of `if let None = ...`

kamalmarhubi:
Make clippy happier.


overvenus:
Thanks @kamalmarhubi 

PTAL @siddontang

ngaut:
LGTM

PR #144, opened 2018-01-25
Ignore clippy build failure

overvenus:
This PR adds an allow-failed build job and enables clippy lint on the job only. Now, annoy compile errors will not fail the whole ci.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,15 +22,22 @@ env:
     - FEATURES="push process"
     - ARCH=i686
 
+matrix:
+  fast_finish: true
+  include:
+  - env: LINT="dev"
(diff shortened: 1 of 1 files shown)


breeswish:
Travis CI infrastructure meets an issue these days. I think we can ignore osx builds for now.

breeswish:
LGTM

breeswish:
Since clippy requires latest Rust nightly to work, I think we can either pin the rust nightly version, or unpin the clippy version (i.e. exclude from cargo) to make them always sync. How do you think? @overvenus

breeswish:
Oh, never mind. Our cargo.lock in fact is not committed.

PR #153, opened 2018-02-07
counter: private LocalCounter::new

overvenus:
`LocalCounter` should not export `new`, it can only be derived from a `Counter`.


UncP:
LGTM

PR #160, opened 2018-03-14
expose memory metrics

UncP:


UncP:
https://github.com/pingcap/rust-prometheus/pull/160

PR #164, opened 2018-04-10
Improve documentation

breeswish:
Previously, the underlying implementation like `GenericCounter`, `GenericCounterVec` are not exposed to the user, so that users cannot know the interface of `(Local)(Int)Counter(Vec)` and `(Int)Gauge(Vec)` totally from the documentation, which is very bad. This PR fixes it.

In addition, this PR improves links for crate structs.


src/lib.rs
```diff
@@ -61,10 +61,37 @@ mod atomic64;
...
+
+    */
+
+    // TODO: define a trait and implement the trait for metrics.
+
+    pub use super::counter::{LocalCounter, LocalCounterVec, LocalIntCounter, LocalIntCounterVec};
+    pub use super::histogram::{LocalHistogram, LocalHistogramTimer, LocalHistogramVec};
+}
+
+pub mod core {
```
overvenus: Why expose these?

Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [package]
 name = "prometheus"
-version = "0.4.0"
+version = "0.4.1"
```
overvenus: Let's save it util next release.

src/histogram.rs
```diff
@@ -445,16 +445,16 @@ impl MetricVecBuilder for HistogramVecBuilder {
...
+/// same [`Desc`](::core::Desc), but have different values for their variable labels. This is used
 /// if you want to count the same thing partitioned by various dimensions
 /// (e.g. HTTP request latencies, partitioned by status code and method).
 pub type HistogramVec = MetricVec<HistogramVecBuilder>;
 
 impl HistogramVec {
-    /// `new` creates a new `HistogramVec` based on the provided `HistogramOpts` and
-    /// partitioned by the given label names. At least one label name must be
-    /// provided.
+    /// Create a new [`HistogramVec`](::core::HistogramVec) based on the provided
```
overvenus: I just found `http:/.../prometheus/::core::HistogramVec` is broken.

breeswish: You need latest nightly rust. This syntax is just landed, see https://github.com/rust-lang/rust/issues/43466

src/lib.rs
breeswish: Otherwise their interface can't be seen in the documentation. I moved those referenced but commonly-won't-use things into the `core` mod.

src/histogram.rs
BusyJay: So which rustc is used on docs.rs?

breeswish: According to [its about page](https://docs.rs/about), it uses the nightly release. Hope it works!

overvenus: It looks like a wrong path, `HistogramVec` does not belong to `core`. Also, please make sure that it does break other links.

breeswish: Oh sorry, it's a mistake. Fixed.

src/counter.rs
```diff
@@ -220,13 +230,18 @@ impl<P: Atomic> Clone for GenericLocalCounter<P> {
...
 }
 
+/// The underlying implementation for [`LocalCounterVec`](::local::LocalCounterVec)
+/// and [`LocalIntCounterVec`](::local::LocalIntCounterVec).
 pub struct GenericLocalCounterVec<P: Atomic> {
     vec: GenericCounterVec<P>,
     local: HashMap<u64, GenericLocalCounter<P>>,
 }
 
+/// An unsync thread local copy of [`CounterVec`](::CounterVec).
```
overvenus: I think `thead local` is a little bit misleading.

breeswish: agree. I think `unsync` is enough?

overvenus: Yes, I think `unsync` is enough too.

PR #173, opened 2018-05-11
Fix protobuf dependency, as 1.4.1 makes builds fail.

### LINKED ISSUE #172: The crate fails to build after the update.
The build log and relevant part of Cargo.lock is in the following gist: https://gist.github.com/niksaak/39d0af30ff04ebe08afb80cbb4275307

niksaak:
Fixes #172.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ name = "benches"
 path = "benches/benches.rs"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


breeswish:
Thanks! As a hot fix I'm accepting this PR but I prefer to fixing the build failure instead of rollbacking the dependency version.

PR #175, opened 2018-05-22
Rearrange benches

overvenus:
This rearranges benches, also fixes a warning: 

```
warning: An explicit [[bench]] section is specified in Cargo.toml which currently
disables Cargo from automatically inferring other benchmark targets.
```


breeswish:
LGTM

PR #182, opened 2018-06-19
Make compile-time-unknown static metrics type safe

breeswish:
# RFC

## Background

When applying compile-time-unknown static metrics, currently we follow this style:

```rust
// Suppose we have `let label = "post"` elsewhere.
SOME_STATIC_VEC_METRIC.get(label).unwrap().inc();
```

As an equivalent to:

```rust
SOME_STATIC_VEC_METRIC.post.inc();
```

Since the value of `label` is unknown in compile time, we have to match it into the actual field at runtime, by using `.get(label).unwrap()`.

According to benchmark, this does not introduce notable overheads, which is good. However this cannot ensure safety: what if `label`'s value is not valid? This will lead to `unwrap()` panic.

From existing applications, we found out that in fact in most of the time we can make it safe in the compile time! We can simply assign an `enum` type for it.

## Design

For `label_enum` definitions, we expand it into real Rust `enum`s so that it can be used in the code elsewhere as a type-safe one:

```rust
make_static_metric! {
    pub label_enum Methods {
        post,
        get,
        put,
        delete,
    }
}
```

will be expand to:

```rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub enum Methods {
    post,
    get,
    put,
    delete,
}
```

Next, we introduce our new `get()` function for metrics.

Take this definition as an example:

```rust
make_static_metric! {
    pub label_enum Methods {
        post,
        get,
        put,
        delete,
    }

    pub struct MyStaticCounterVec: Counter {
        "method" => Methods,
    }
}
```

The `get()` function of `MyStaticCounterVec` will be expanded as:

```rust
impl MyStaticCounterVec {
    pub fn get(&self, label: Methods): Counter {
        match label {
            Methods::post => self.post,
            Methods::get => self.get,
            Methods::put => self.put,
            Methods::delete => self.delete,
        }
    }
}
```

The old `get()` function will be renamed to `try_get` and other signatures remain unchanged.

If the label is provided inline like:

```rust
make_static_metric! {
    pub struct MyAnotherStaticCounterVec: IntCounter {
        "error" => {
            error_1,
            error_2,
        },
    }
}
```

`get()` will not be available and no enums will be generated accordingly.

---

During developing, I found out that some names need to be refined, otherwise there will be confusions. So this PR contains pretty much changes. However I organized my changes in commits so that you can review by commits, which is very friendly.

[Commit 1](https://github.com/pingcap/rust-prometheus/commit/c145f5c4186f564285461c49890634a0c565d27d) implements the RFC and is working (see example in the commit), exclude the derive part (see commit 4).
[Commit 2](https://github.com/pingcap/rust-prometheus/commit/a54af54c86d9f75225fc75efb16d8d79c4157feb) refines some misleading and confusing names in the parser to make it more clear. This commit does not provide new features.
[Commit 3](https://github.com/pingcap/rust-prometheus/commit/48a20d67b8ec210b27bf03040d9c8d26580e2621) adjusts acceptable expressions for the label value. Previously we accept `Expr`, which is any Rust expression. However in fact only string literals will work according to our implementations. So this commit changed `Expr` to `LitStr` to restrict the type.
[Commit 4](https://github.com/pingcap/rust-prometheus/commit/c9aa33ace2602d15b12065c5acfd656ffa7f52d4) implements Clone, Copy, Debug, Display, etc. for the enum.
[Commit 5](https://github.com/pingcap/rust-prometheus/commit/22feeb1855172f0eb451fcbbb9e3767b76d56e94) adds tests to further ensure stability and quality.

### DIFF
--- a/static-metric/examples/metric_enum.rs
+++ b/static-metric/examples/metric_enum.rs
@@ -26,14 +26,14 @@ use prometheus::{CounterVec, IntCounterVec, Opts};
 use prometheus_static_metric::make_static_metric;
 
 make_static_metric! {
-    label_enum Methods {
+    pub label_enum Methods {
         post,
         get,
         put,
         delete,
     }
 
-    label_enum Products {
+    pub label_enum Products {
         foo,
         bar,
     }
@@ -61,6 +61,10 @@ fn main() {
     assert_eq!(static_counter_vec.post.bar.get(), 0.0);
     assert_eq!(vec.with_label_values(&["post", "foo"]).get(), 1.0);
     assert_eq!(vec.with_label_values(&["delete", "bar"]).get(), 4.0);
+
+    // metric enums will expose an enum for type-safe `get()`.
+    static_counter_vec.get(Methods::post).foo.inc();
+    assert_eq!(static_counter_vec.post.foo.get(), 2.0);
 
     let vec = IntCounterVec::new(Opts::new("foo", "bar"), &["error", "error_method"]).unwrap();
     let static_counter_vec = MyAnotherStaticCounterVec::from(&vec);
--- a/static-metric/Cargo.toml
+++ b/static-metric/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-static-metric"
-version = "0.1.2"
+version = "0.1.3"
 license = "Apache-2.0"
 authors = ["me@breeswish.org"]
 description = "Static metric helper utilities for rust-prometheus."
--- a/static-metric/examples/advanced.rs
+++ b/static-metric/examples/advanced.rs
@@ -70,9 +70,9 @@ fn main() {
     // Note: You cannot specify values other than the definition in `get()` because
     // it is purely static.
     HTTP_COUNTER
-        .get("delete")
+        .try_get("delete")
         .unwrap()
-        .get("HTTP/1")
+        .try_get("HTTP/1")
         .unwrap()
         .foo
         .inc_by(7);
--- a/static-metric/src/builder.rs
+++ b/static-metric/src/builder.rs
@@ -108,9 +108,9 @@ impl TokensBuilder {
 
             #[allow(dead_code)]
             mod #scope_name {
-                use std::collections::HashMap;
-                use prometheus::#metric_type;
-                use prometheus::#metric_vec_type;
+                use ::std::collections::HashMap;
+                use ::prometheus::#metric_type;
+                use ::prometheus::#metric_vec_type;
 
                 #[allow(unused_imports)]
                 use super::*;
@@ -125,12 +125,12 @@ impl TokensBuilder {
     fn build_label_enum(label_enum: &MetricEnumDef) -> Tokens {
         let visibility = &label_enum.visibility;
         let enum_name = &label_enum.enum_name;
-        let enum_values = label_enum.get_values().iter().map(|value| value.name);
+        let enum_item_names = label_enum.definitions.get_names();
         quote!{
             #[allow(dead_code)]
             #[allow(non_camel_case_types)]
             #visibility enum #enum_name {
-                #(#enum_values),*
+                #(#enum_item_names),*
             }
         }
     }
@@ -176,11 +176,9 @@ impl<'a> MetricBuilderContext<'a> {
     fn build_struct(&self) -> Tokens {
         let struct_name = &self.struct_name;
 
-        let field_names: Vec<_> = self.label
-            .get_values(self.enum_definitions)
-            .iter()
-            .map(|v| &v.name)
-            .collect();
+        let field_names = self.label
+            .get_value_def_list(self.enum_definitions)
+            .get_names();
         let member_types: Vec<_> = field_names.iter().map(|_| &self.member_type).collect();
 
         quote!{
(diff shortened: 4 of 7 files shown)


breeswish:
This PR expands the following:

```rust
make_static_metric! {
    pub label_enum Methods {
        post,
        get,
        put,
        delete,
    }

    pub label_enum Products {
        foo,
        bar,
    }

    pub struct MyStaticCounterVec: Counter {
        "method" => Methods,
        "product" => Products,
    }
}
```

into:

```rust
#[allow(dead_code)]
#[allow(non_camel_case_types)]
#[derive(Clone, Copy, PartialEq)]
pub enum Methods {
    post,
    get,
    put,
    delete,
}
impl ::std::fmt::Debug for Methods {
    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
        match self {
            Methods::post => write!(f, "post"),
            Methods::get => write!(f, "get"),
            Methods::put => write!(f, "put"),
            Methods::delete => write!(f, "delete"),
        }
    }
}
impl ::std::fmt::Display for Methods {
    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
        ::std::fmt::Debug::fmt(self, f)
    }
}
#[allow(dead_code)]
#[allow(non_camel_case_types)]
#[derive(Clone, Copy, PartialEq)]
pub enum Products {
    foo,
    bar,
}
impl ::std::fmt::Debug for Products {
    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
        match self {
            Products::foo => write!(f, "foo"),
            Products::bar => write!(f, "bar"),
        }
    }
}
impl ::std::fmt::Display for Products {
    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
        ::std::fmt::Debug::fmt(self, f)
    }
}
pub use self::prometheus_static_scope_0::MyStaticCounterVec;
#[allow(dead_code)]
mod prometheus_static_scope_0 {
    #[allow(unused_imports)]
    use super::*;
    use prometheus::Counter;
    use prometheus::CounterVec;
    #[allow(missing_copy_implementations)]
    pub struct MyStaticCounterVec {
        pub post: MyStaticCounterVec2,
        pub get: MyStaticCounterVec2,
        pub put: MyStaticCounterVec2,
        pub delete: MyStaticCounterVec2,
    }
    impl MyStaticCounterVec {
        pub fn from(m: &CounterVec) -> MyStaticCounterVec {
            MyStaticCounterVec {
                post: MyStaticCounterVec2::from("post", m),
                get: MyStaticCounterVec2::from("get", m),
                put: MyStaticCounterVec2::from("put", m),
                delete: MyStaticCounterVec2::from("delete", m),
            }
        }
        pub fn get(&self, value: Methods) -> &MyStaticCounterVec2 {
            match value {
                Methods::post => &self.post,
                Methods::get => &self.get,
                Methods::put => &self.put,
                Methods::delete => &self.delete,
            }
        }
        pub fn try_get(&self, value: &str) -> Option<&MyStaticCounterVec2> {
            match value {
                "post" => Some(&self.post),
                "get" => Some(&self.get),
                "put" => Some(&self.put),
                "delete" => Some(&self.delete),
                _ => None,
            }
        }
    }
    #[allow(missing_copy_implementations)]
    pub struct MyStaticCounterVec2 {
        pub foo: Counter,
        pub bar: Counter,
    }
    impl MyStaticCounterVec2 {
        pub fn from(label_0: &str, m: &CounterVec) -> MyStaticCounterVec2 {
            MyStaticCounterVec2 {
                foo: m.with(&{
                    let mut coll = HashMap::new();
                    coll.insert("method", label_0);
                    coll.insert("product", "foo");
                    coll
                }),
                bar: m.with(&{
                    let mut coll = HashMap::new();
                    coll.insert("method", label_0);
                    coll.insert("product", "bar");
                    coll
                }),
            }
        }
        pub fn get(&self, value: Products) -> &Counter {
            match value {
                Products::foo => &self.foo,
                Products::bar => &self.bar,
            }
        }
        pub fn try_get(&self, value: &str) -> Option<&Counter> {
            match value {
                "foo" => Some(&self.foo),
                "bar" => Some(&self.bar),
                _ => None,
            }
        }
    }
}
```

breeswish:
@overvenus @Hoverbear PTAL

Hoverbear:
@breeswish In the description you reference an RFC, can you please link to it?

static-metric/examples/metric_enum.rs
```diff
@@ -62,6 +62,10 @@ fn main() {
     assert_eq!(vec.with_label_values(&["post", "foo"]).get(), 1.0);
     assert_eq!(vec.with_label_values(&["delete", "bar"]).get(), 4.0);
 
+    // metric enums will expose an enum for type-safe `get()`.
+    static_counter_vec.get(Methods::post).foo.inc();
+    assert_eq!(static_counter_vec.post.foo.get(), 2.0);
```
Hoverbear: Can `static_counter.post.foo.inc()` also work?

breeswish:
@Hoverbear You can take a look at our internal Confluence :)

static-metric/examples/metric_enum.rs
breeswish: yes! This still work, see line 67. This PR enhances type safety for compile-time-unknown labels.

breeswish:
@Hoverbear The RFC was in our internal wiki. I updated my thread and pasted it here as well.

static-metric/examples/metric_enum.rs
Hoverbear: :heart_eyes: Awesome.

Issue #176, opened 2018-05-23
Use `cargo clippy` instead of the clippy plugin


overvenus:
The clippy team is planning on removing plugin interface. See more https://github.com/pingcap/tikv/issues/3090.

Also the `dev` feature should be remove after that.

PR #184, opened 2018-06-20
Use `cargo clippy` instead of the clippy plugin #176

niedhui:
Hi
first sorry for my messy commit message... I did not realized even I use force push to make the commit clean, but leaved the useless history in the issue page.

Currently the clippy result is same as using clippy plugin, except that even there is some clippy issues , the returned status is still 0, maybe related to https://github.com/rust-lang-nursery/rust-clippy/issues/1209, so the CI will mark it as success.


Cargo.toml
```diff
@@ -15,7 +15,7 @@ travis-ci = { repository = "pingcap/rust-prometheus" }
 
 [features]
 default = []
-dev = ["clippy"]
+dev = []
```
Hoverbear: You can remove this feature now. :)

src/lib.rs
```diff
@@ -15,11 +15,9 @@
 The Rust client library for [Prometheus](https://prometheus.io/).
 */
 
-#![cfg_attr(feature = "dev", feature(plugin))]
-#![cfg_attr(feature = "dev", plugin(clippy))]
-#![cfg_attr(not(feature = "dev"), allow(unknown_lints))]
-#![cfg_attr(feature = "dev", allow(needless_pass_by_value))]
-#![cfg_attr(feature = "dev", allow(new_without_default_derive))]
+#![allow(unknown_lints)]
```
Hoverbear: Can you combine these together?

Cargo.toml
niedhui: Ok.  is it safe to also remove these two lines on the Makefile? I had no idea since I had not use any Makefile yet😂
https://github.com/pingcap/rust-prometheus/blob/master/Makefile#L11

.travis.yml
```diff
@@ -25,20 +25,17 @@ env:
...
 script:
   # Test common features.
   - cargo test --features="$FEATURES"
   - |
-    if [ $TRAVIS_RUST_VERSION = 'nightly' ]; then
-        # Enable dev for clippy linting.
-        cargo test --features="$FEATURES $LINT";
-        # Lint nightly feature.
-        cargo test --features="$FEATURES $LINT nightly";
+    if [ $CLIPPY = 'yes' ]; then
```
breeswish: Since CIPPY is always `yes`, can we remove this condition test?

src/lib.rs
breeswish: Previously we allow unknown_lints when feature is not dev. I guess for now this allow should be removed?

Cargo.toml
breeswish: You can remove the dev feature and make `make dev` to be `format + test`.

.travis.yml
niedhui: the CLIPPY env is only  set yes when rust is nightly on the matrix section,  so I thought the condition test is still need

src/lib.rs
niedhui: done, by using 
```
#![cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value, new_without_default_derive))]
```

Cargo.toml
niedhui: done

rustfmt.toml
```diff
@@ -1,5 +1,2 @@
 reorder_imports = true
-reorder_imports_in_group = true
```
breeswish: Why are these lint rules removed?

.travis.yml
breeswish: Oh. Make sense.

rustfmt.toml
breeswish: Oh, I see. These configurations are removed from rustfmt.

breeswish:
LGTM

PR #200, opened 2018-09-15
Provide Histogram::new and support different types for name and help

breeswish:

This PR provides `Histogram::new()`, just like there is `Counter::new()` and `Gauge::new()`.

Also, this PR makes `name` and `help` parameter able to accept different types.


src/counter.rs
```diff
@@ -47,7 +47,11 @@ impl<P: Atomic> Clone for GenericCounter<P> {
 
 impl<P: Atomic> GenericCounter<P> {
     /// Create a [`GenericCounter`](::core::GenericCounter) with the `name` and `help` arguments.
-    pub fn new<S: Into<String>>(name: S, help: S) -> Result<Self> {
```
Hoverbear: You can just do `name: impl Into<String>` now on stable. :)

PR #202, opened 2018-09-15
Provide (un)register functions for collectors

breeswish:

Based on https://github.com/pingcap/rust-prometheus/pull/201

This PR provides `(try)_(un)register` for all collectors, so that the syntax looks nice even without macros. Macros will be fully removed after `From<Opts>` for collectors is implemented.

This PR also extracts common traits (Collector and Encoder) to `prometheus::prelude` module, so that user can easily import them via `use prometheus::prelude::*`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,15 +44,15 @@ features = ["gen"]
 ## Example
 
 ```rust
-use prometheus::{Opts, Registry, Counter, TextEncoder, Encoder};
+use prometheus::prelude::*;
+use prometheus::{Opts, Registry, Counter, TextEncoder};
 
-// Create a Counter.
-let counter_opts = Opts::new("test_counter", "test counter help");
-let counter = Counter::with_opts(counter_opts).unwrap();
-
-// Create a Registry and register Counter.
+// Create a Registry.
 let r = Registry::new();
-r.try_register(Box::new(counter.clone())).unwrap();
+
+// Create a Counter and register to the registry.
+let counter_opts = Opts::new("test_counter", "test counter help");
+let counter = Counter::with_opts(counter_opts).unwrap().register(r);
 
 // Inc.
 counter.inc();
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ let counter = Counter::with_opts(counter_opts).unwrap();
 
 // Create a Registry and register Counter.
 let r = Registry::new();
-r.register(Box::new(counter.clone())).unwrap();
+r.try_register(Box::new(counter.clone())).unwrap();
 
 // Inc.
 counter.inc();
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -28,6 +28,86 @@ pub trait Collector: Sync + Send {
 
     /// Collect metrics.
     fn collect(&self) -> Vec<proto::MetricFamily>;
+
+    /// Alias of [`Registry::try_register`]. Register the collector to a registry.
+    fn try_register(self, registry: &super::Registry) -> Result<Self>
+    where
+        Self: 'static + Sized + Clone,
+    {
+        registry.try_register(Box::new(self.clone())).map(|_| self)
+    }
+
+    /// Alias of [`Registry::register`]. Register the collector to a registry.
+    /// Panics if there are errors.
+    fn register(self, registry: &super::Registry) -> Self
+    where
+        Self: 'static + Sized + Clone,
+    {
+        registry
+            .try_register(Box::new(self.clone()))
+            .map(|_| self)
+            .unwrap()
+    }
+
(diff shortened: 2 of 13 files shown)


src/metrics.rs
```diff
@@ -28,6 +28,86 @@ pub trait Collector: Sync + Send {
...
+    where
+        Self: 'static + Sized + Clone,
+    {
+        registry
+            .try_unregister(Box::new(self.clone()))
+            .map(|_| self)
+    }
+
+    /// Alias for [`Registry::try_unregister`]. Unregister the collector from a registry.
+    /// Panics if there are errors.
```
Hoverbear: Ditto to above.

```
# Panics

* Reasons why it panics
```

src/registry.rs
```diff
@@ -31,7 +31,7 @@ struct RegistryCore {
 }
 
 impl RegistryCore {
-    fn register(&mut self, c: Box<Collector>) -> Result<()> {
+    fn try_register(&mut self, c: Box<Collector>) -> Result<()> {
```
Hoverbear: I think this is unnecessary, returning `Result<()>` already implies a try.

src/registry.rs
```diff
@@ -88,7 +88,7 @@ impl RegistryCore {
         }
     }
 
-    fn unregister(&mut self, c: Box<Collector>) -> Result<()> {
+    fn try_unregister(&mut self, c: Box<Collector>) -> Result<()> {
```
Hoverbear: Ditto.

src/registry.rs
```diff
@@ -207,12 +207,12 @@ impl Default for Registry {
...
 
 impl Registry {
-    /// `new` creates a Registry.
+    /// Create a Registry.
     pub fn new() -> Registry {
         Registry::default()
     }
 
-    /// `register` registers a new [`Collector`](::core::Collector) to be included in metrics
+    /// Register a new [`Collector`](::core::Collector) to be included in metrics
```
Hoverbear: If you make a line, then some more content, rustdoc will show the first line as a summary on doc pages. Example: https://docs.rs/raft/0.3.1/raft/#structs

```
Register a new `Collect` to be included in metric collection.

It returns...
```

src/registry.rs
```diff
@@ -221,16 +221,28 @@ impl Registry {
...
     /// Describe method yields the same set of descriptors.) The function
     /// returns error when the [`Collector`](::core::Collector) is not registered.
-    pub fn unregister(&self, c: Box<Collector>) -> Result<()> {
-        self.r.write().unregister(c)
+    pub fn try_unregister(&self, c: Box<Collector>) -> Result<()> {
+        self.r.write().try_unregister(c)
+    }
+
+    /// Unregister the [`Collector`](::core::Collector) that equals the [`Collector`](::core::Collector) passed
+    /// in as an argument. Panics if there are errors.
```
Hoverbear: Please use a panics section:

```
# Panics

* Reasons it might panic.
```

src/registry.rs
```diff
@@ -268,20 +280,25 @@ lazy_static! {
...
-/// do not fulfill the consistency and uniqueness criteria described in the [`Desc`](::core::Desc)
-/// documentation.
-pub fn register(c: Box<Collector>) -> Result<()> {
+/// Registers a new [`Collector`](::core::Collector) to the default registry.
+pub fn try_register(c: Box<Collector>) -> Result<()> {
+    DEFAULT_REGISTRY.try_register(c)
+}
+
+/// Registers a new [`Collector`](::core::Collector) to the default registry, panics if
+/// there are errors.
```
Hoverbear: Panic section here too. :)

README.md
```diff
@@ -44,15 +44,15 @@ features = ["gen"]
 ## Example
 
 ```rust
-use prometheus::{Opts, Registry, Counter, TextEncoder, Encoder};
+use prometheus::prelude::*;
+use prometheus::{Opts, Registry, Counter, TextEncoder};
```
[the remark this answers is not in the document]
kennytm: In https://docs.rs/futures/0.1/futures/prelude/index.html, `Async` and `AsyncSink` are enums, and `Poll` and `StartSend` are type aliases 🙃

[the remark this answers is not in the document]
breeswish: so it does not conflict with "mostly traits" :) In addition, the link given is a clear use case that we need prelude as well as other imports.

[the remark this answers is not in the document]
Hoverbear: Ok. :)

Hoverbear:
Please see proposal to close #201. While I like `unregister`, I really don't like having the default API panic.

breeswish:
Will open another one based on comments. Closing.

PR #205, opened 2018-10-01
Use std is_ascii

zkanda:
Fixes this todo: https://github.com/pingcap/rust-prometheus/blob/a5a2d7c3d3c167a6da518b71a6c9212cb6b8f3bf/src/desc.rs#L24

### DIFF
--- a/src/desc.rs
+++ b/src/desc.rs
@@ -21,26 +21,19 @@ use errors::{Error, Result};
 use metrics::SEPARATOR_BYTE;
 use proto::LabelPair;
 
-// TODO: use `char::is_ascii` instead once it landed in the stable rust.
-// Refer to https://github.com/rust-lang/rust/blob/
-//          3e9a7f7fbbf2898b9f1d60886f92e76370040d83/src/libstd_unicode/char.rs#L943
-fn is_ascii(c: char) -> bool {
-    c as u32 <= 0x7F
-}
-
 // Details of required format are at
 //   https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
 fn is_valid_metric_name(name: &str) -> bool {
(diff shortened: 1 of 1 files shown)


src/desc.rs
```diff
@@ -21,26 +21,19 @@ use errors::{Error, Result};
...
-    c as u32 <= 0x7F
-}
-
 // Details of required format are at
 //   https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
 fn is_valid_metric_name(name: &str) -> bool {
     // Valid metric names must match regex [a-zA-Z_:][a-zA-Z0-9_:]*.
     fn valid_start(c: char) -> bool {
-        is_ascii(c) && match c as u8 {
+        c.is_ascii() && match c as u8 {
```
kennytm: 🤔 Have we ever checked if doing this `is_ascii` check is faster than the more straightforward

```rust
match c {
    'a'..='z' | 'A'..='Z' | '_' | ':' => true,
    _ => false,
}
```

breeswish: Good question! Also valid metric name should be more common so it will be nice if valid metric name is the fast path.

```rust
fn is_valid(c: char) -> bool {
    c.is_ascii() && match c as u8 {
        b'a'...b'z' | b'A'...b'Z' | b'_' | b':' => true,
        _ => false,
    }
}

fn is_valid_simple(c: char) -> bool {
    match c {
        'a'..='z' | 'A'..='Z' | '_' | ':' => true,
        _ => false,
    }
}
```

I made a simple benchmark, and seems that `is_valid_simple` is very stable, which always takes 0ns while `is_valid` may take `1ns` or `0ns` for different inputs.

zkanda: thanks for benchmarking it @breeswish I guess we can close this PR since the simple one is better. 🍡

PR #219, opened 2019-01-17
Update hyper to 0.12

remram44:
This updates from [0.9](https://github.com/pingcap/rust-prometheus/pull/17) to 0.12, based on [this example](https://docs.rs/hyper/0.12.21/hyper/server/index.html#example).

PR #223, opened 2019-02-06
Fix linting

remram44:
Let's fix linting on master before tackling #219 :wink:


breeswish:
Why there is 0 change for this PR now 😅

remram44:
@fralalonde snuck in the same changes with 89ca6991 :laughing:

PR #225, opened 2019-02-21
Fix build for static-metric benches

brson:
The `get` accessors in the current code don't resolve. From my reading, it looks like this macro invocation is in the proper form to produce such accessors.

PTAL @breeswish


brson:
This is old and I don't know if it's still relevant. Closing.

PR #220, opened 2019-01-28
Rust 2018

brson:
This upgrades prometheus to 2018 and bumps the version from 0.5 to 0.6.

Unfortunately I haven't been able to test this against tikv/master: tikv is currently on prometheus 0.4, and 0.5 added a dependency on openssl that doesn't link with tikv for mysterious reasons that I haven't investigated. Still, i don't have much expectation that just upgrading to 2018 will cause incomptabilities with tikv's current toolchain.

This will conflict a lot with https://github.com/pingcap/rust-prometheus/pull/216, so it might be appropriate to get that landed and let me deal with rebasing.

Fixes https://github.com/pingcap/rust-prometheus/issues/217

PTAL @overvenus

PR #226, opened 2019-03-06
Use the 2018 edition

nrc:
Also fixes up the Clippy attribute and removes rustfmt.toml


kennytm:
Perhaps this should be combined into  #220?

nrc:
> Perhaps this should be combined into #220?

Ah, yep, should have checked the open PRs.

nrc:
Closing in favour of #220 and #227

PR #227, opened 2019-03-08
Support Prost

nrc:
~~This can't be merged yet because it relies on path dependencies (waiting on changes to land in Prost)~~. Based on https://github.com/pingcap/rust-prometheus/pull/226


nrc:
This is now ready for review. PTAL @overvenus

overvenus:
Hi @nrc There are few warnings, would you mind address them? https://travis-ci.org/pingcap/rust-prometheus/jobs/518986584

nrc:
Hi addressed the warnings and rebased, but in the course of rebasing it became a bit more tricky to support both Prost and rust-protobuf. Since that is just a 'nice to have' for the migration, I've removed rust-protobuf support. I also removed support for having no codec, since I'm not sure why you would ever build without Prost (there is no version issue and protoc is not required)

src/encoder/pb.rs
```diff
@@ -38,10 +38,12 @@ impl ProtobufEncoder {
 
 impl Encoder for ProtobufEncoder {
     fn encode<W: Write>(&self, metric_families: &[MetricFamily], writer: &mut W) -> Result<()> {
+        let mut buf = Vec::new();
```
nrc: For the sake of performance, we should change `encode` to take a `Bytes` buffer rather than a `Write` object, but I'm not sure how big a breaking change that would be.

nrc: @overvenus what do you think?

nrc: ping @overvenus  do you have thoughts on this?

breeswish: I think we can leave it since ProtoBuf format will be removed in future (Prometheus has already dropped support for it).

src/vec.rs
```diff
@@ -48,14 +48,14 @@ impl<T: MetricVecBuilder> MetricVecCore<T> {
         let mut m = MetricFamily::default();
         m.set_name(self.desc.fq_name.clone());
         m.set_help(self.desc.help.clone());
-        m.set_field_type(self.metric_type);
+        m.set_field_type_(self.metric_type);
```
breeswish: Why is there a `_` suffix?

breeswish:
I'm fine with this PR. However I would like to see 0.6 carrying https://github.com/pingcap/rust-prometheus/pull/235 https://github.com/pingcap/rust-prometheus/pull/237 and https://github.com/pingcap/rust-prometheus/pull/238. Maybe leave this PR to 0.7 is better.

src/encoder/pb.rs
nrc: :+1:

src/vec.rs
nrc: For enums, the `set_field_type` method is already generated by Prost, so we need to use a different name for the version we generate.

nrc:
@breeswish is it OK to merge this PR now?

breeswish:
Looks good. Could you fix the conflict?

Cargo.toml
```diff
@@ -17,31 +18,34 @@ travis-ci = { repository = "pingcap/rust-prometheus" }
...
-default = ["protobuf"]
 nightly = ["libc"]
 push = ["reqwest", "libc"]
 process = ["libc", "procinfo"]
-gen = ["protobuf-codegen-pure"]
+gen = ["protobuf-build"]
 
 [dependencies]
-protobuf = { version = "2.0", optional = true }
+protobuf = "2.0"
```
breeswish: Can you keep it optional? See https://github.com/pingcap/rust-prometheus/pull/216

nrc:
Hmm, after reading #216, it seems we don't need Prost support in Prometheus, so I think I can just close this PR.

Issue #230, opened 2019-03-17
Expose the default registry


lucab:
Context: I'm currently trying to instrument library and application logic in an ergonomic way, see https://github.com/pingcap/rust-prometheus/issues/228 for an example.

I'm playing with the idea of structuring my library logic so that applications can inject their own `Registry` at initialization time. However some of them are currently using the default registry, and I don't see a way to get a handle to it (did I maybe miss it?).

As a consumer, I think I would like to:
 * have a `pub fn default_registry() -> &Registry`
 * ensure it internally initializes the registry (see https://docs.rs/lazy_static/1.3.0/lazy_static/fn.initialize.html)
 * slowly deprecate top-level `register`/`unregister`/`gather` in favor of registry-scoped methods

PR #231, opened 2019-03-25
prometheus: expose the default registry

lucab:
This adds a getter to expose the default/static registry. It also includes an example showing how to use multiple and custom registries, and how to perform registration across function boundaries.

Closes: https://github.com/pingcap/rust-prometheus/issues/230

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -268,6 +268,12 @@ lazy_static! {
     };
 }
 
+/// Default registry (global static).
+pub fn default_registry() -> &'static Registry {
+    lazy_static::initialize(&DEFAULT_REGISTRY);
+    &DEFAULT_REGISTRY
+}
+
 /// Registers a new [`Collector`](::core::Collector) to be included in metrics collection. It
 /// returns an error if the descriptors provided by the [`Collector`](::core::Collector) are invalid or
 /// if they - in combination with descriptors of already registered Collectors -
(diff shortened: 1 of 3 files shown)


lucab:
Temporarily rebased on top of https://github.com/pingcap/rust-prometheus/pull/232 to avoid Travis failure.

src/registry.rs
```diff
@@ -268,6 +268,12 @@ lazy_static! {
     };
 }
 
+/// Default registry (global static).
+pub fn default_registry() -> &'static Registry {
+    lazy_static::initialize(&DEFAULT_REGISTRY);
```
breeswish: Or can we just deref the static created by lazy static?

lucab: We just need to somehow ensure the Registry is initialized before handing it out, so yes we can.
That would a be a
```rust
&*DEFAULT_REGISTRY;
```
which I find a little less intuitive to understand than the current form (and I'm a bit worried about dead-code optimization), but could be explained with a comment.
Can we stick to the current form (my preference) or do you want me to swap with the one above?

breeswish: Both are fine. I was used to the `&*` form :)

breeswish:
@overvenus PTAL

src/registry.rs
lucab: Ack. I'll keep the explicit `initialize` then.

lucab:
I just rebased on current master and force-pushed to get the Travis fix.

breeswish:
Lose contact with @overvenus . I'm going to merge this PR because it looks fine!

breeswish:
Thanks a lot!

lucab:
Do we want to add a deprecation notice on top-level `register`/`unregister`/`gather` or do you prefer to leave them as-is?

Issue #228, opened 2019-03-15
registry: namespacing all registered collectors/metrics


lucab:
It would be nice to have an additional `prefix` field stored within `Registry`, so that all collectors/metrics registered with it are automatically namespaced with the relevant prefix (when encoding).

The usecase for this is shared/library logic recording internal metrics that are exposed by consuming applications. That is:
 * library counter: `outgoing_requests_total`
 * app0 consumer:
   - registry prefix: `app0`
   - metric name: `app0_outgoing_requests_total`
* app1 consumer:
   - registry prefix: `app1`
   - metric name: `app1_outgoing_requests_total`

This would mimic client-go `WrapRegistererWithPrefix`: https://godoc.org/github.com/prometheus/client_golang/prometheus#WrapRegistererWithPrefix

lucab:
As a side-note, I think there are similar usecases for registry-wide `labels` too.

breeswish:
Good suggestion! Since golang client support this, we should consider this! I will do it when I'm free. Feel free to propose a PR before me.

PR #232, opened 2019-03-25
clippy: fix renamed lint

lucab:
This fixes a clippy hard-failure due to a renamed lint:
```
error: lint `clippy::new_without_default_derive` has been renamed to `clippy::new_without_default`
```

PR #233, opened 2019-03-25
registry: add optional common prefix and labels

lucab:
This adds an additional constructor to provide common namespace/labels
for a registry, to be used for all collected metrics.

Closes: https://github.com/pingcap/rust-prometheus/issues/228

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -15,7 +15,6 @@
 use std::collections::btree_map::Entry as BEntry;
 use std::collections::hash_map::Entry as HEntry;
 use std::collections::{BTreeMap, HashMap, HashSet};
-use std::iter::FromIterator;
 use std::sync::Arc;
 
 use spin::RwLock;
@@ -28,6 +27,10 @@ struct RegistryCore {
     pub collectors_by_id: HashMap<u64, Box<Collector>>,
     pub dim_hashes_by_name: HashMap<String, u64>,
     pub desc_ids: HashSet<u64>,
+    /// Optional common labels for all registered collectors.
+    pub labels: Option<HashMap<String, String>>,
+    /// Optional common namespace for all registered collectors.
+    pub prefix: Option<String>,
 }
 
 impl RegistryCore {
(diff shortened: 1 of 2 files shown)


lucab:
This is rebased on top #231 and #232 as it re-uses the same Travis fix and examples. I'll update this PR once those two dependencies are merged.

lucab:
I amended this PR as I've found some time to fit the "common labels" logic too.

lucab:
@overvenus @breeswish rebased on master as the other PRs are in, this is now ready for review.

src/registry.rs
```diff
@@ -179,9 +182,36 @@ impl RegistryCore {
...
+                            label.set_name(k.to_string());
+                            label.set_value(v.to_string());
+                            label
+                        })
+                        .collect();
+
+                    let metrics = m.mut_metric().as_mut_slice();
+                    for entry in metrics {
+                        let labels = entry.mut_label();
+                        for entry in &pairs {
```
breeswish: Maybe we can use [Vec::append](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append) so that you don't need this for loop?

lucab: The underlying collection is a [RepeatedField](https://docs.rs/protobuf/2.2.4/protobuf/struct.RepeatedField.html), I didn't see a method to append-in-place. Maybe I missed it? Or maybe are are you suggesting to go through something like the following?
```rust
for metric in m.mut_metric().iter_mut() {
  let mut labels = metric.take_label().into_vec();
  labels.append(&mut pairs);
  metric.set_label(labels.into());
}
```

breeswish: How about calling `into_vec()`?

lucab: Indeed. I've updated the code to do that.

lucab:
Amended with @breeswish suggestion.

lucab:
@breeswish @overvenus @Hoverbear thanks for reviewing and merging! Can we tag a 0.6.0 release with this?

Issue #236, opened 2019-04-24
Panic in debug due to int overflow in collector_id


jsham:
**Describe the bug**
Running debug build might crash, due to collector_id overflowing (which is intended!)
This can be fixed by using `std::num::Wrapping` type for hashes.

```thread 'tokio-runtime-worker-1' panicked at 'attempt to add with overflow', /builds/..../_cargo/registry/src/github.com-1ecc6299db9ec823/prometheus-0.5.0/src/registry.rs:97:17```

breeswish:
Thanks for your report!

breeswish:
Fixed in #237

PR #237, opened 2019-04-25
Fix collector_id overflow panic

breeswish:
See https://github.com/pingcap/rust-prometheus/issues/236


src/registry.rs
```diff
@@ -93,11 +93,11 @@ impl RegistryCore {
 
     fn unregister(&mut self, c: Box<Collector>) -> Result<()> {
         let mut id_set = Vec::new();
-        let mut collector_id = 0;
+        let mut collector_id: u64 = 0;
         for desc in c.desc() {
             if id_set.iter().find(|id| **id == desc.id).is_none() {
                 id_set.push(desc.id);
-                collector_id += desc.id;
+                collector_id = collector_id.wrapping_add(desc.id);
```
nrc: I think it is better to use `checked_add` and return an Error, rather than silently wrapping around.

breeswish: It is intended to overflow

breeswish:
@overvenus @siddontang PTAL

breeswish:
ping @overvenus @siddontang

nrc:
@breeswish can we merge this PR and release 0.6 now?

PR #240, opened 2019-05-06
Fix compilation without protobuf

meqif:
Compilation of the latest version (0.6.0) failed in one of my projects with the following error:

```
error[E0599]: no method named `take_label` found for type `&mut proto::Metric` in the current scope
   --> src/registry.rs:208:49
    |
208 |                         let mut labels = metric.take_label().into_vec();
    |                                                 ^^^^^^^^^^

```

Upon investigation, I found out this happened if the crate was configured not to use the protobuf feature (e.g. with `default-features = false` when listed as a dependency, or with `cargo build --no-default-features`).

Further investigation revealed that the `Metric` implementation in `plain_model.rs` was missing the `mut_label` and `take_label` methods, which exist in the version defined in `proto/proto_model.rs`. Fixing that was easy enough, but required a small adaptation in `registry.rs`:

```diff
-                         let mut labels = metric.take_label().into_vec();
+                         let mut labels: Vec<_> = metric.take_label().into();
```

I also took the opportunity to fix the deprecation warning about `LabelPair::new`.

Please let me know if there's anything you wish I'd improve in this PR.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -197,7 +197,7 @@ impl RegistryCore {
                     let mut pairs: Vec<proto::LabelPair> = hmap
                         .iter()
                         .map(|(k, v)| {
-                            let mut label = proto::LabelPair::new();
+                            let mut label = proto::LabelPair::default();
                             label.set_name(k.to_string());
                             label.set_value(v.to_string());
                             label
@@ -205,7 +205,7 @@ impl RegistryCore {
                         .collect();
 
                     for metric in m.mut_metric().iter_mut() {
-                        let mut labels = metric.take_label().into_vec();
+                        let mut labels: Vec<_> = metric.take_label().into();
                         labels.append(&mut pairs);
                         metric.set_label(labels.into());
                     }
@@ -506,7 +506,7 @@ mod tests {
         assert_eq!(mfs.len(), 1);
         assert_eq!(mfs[0].get_name(), "test_a_counter");
 
-        let mut needle = proto::LabelPair::new();
+        let mut needle = proto::LabelPair::default();
         needle.set_name("tkey".to_string());
         needle.set_value("tvalue".to_string());
         let metrics = mfs[0].get_metric();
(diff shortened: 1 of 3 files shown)


src/registry.rs
```diff
@@ -197,15 +197,15 @@ impl RegistryCore {
...
+                            let mut label = proto::LabelPair::default();
                             label.set_name(k.to_string());
                             label.set_value(v.to_string());
                             label
                         })
                         .collect();
 
                     for metric in m.mut_metric().iter_mut() {
-                        let mut labels = metric.take_label().into_vec();
+                        let mut labels: Vec<_> = metric.take_label().into();
```
meqif: `Vec` doesn't implement `into_vec`, so I had to go around and use `Into::into`.

breeswish:
Thanks for the report and the PR.

@overvenus Looks like we need to adjust our CI matrix to cover different features.

src/registry.rs
breeswish: How about just `let mut labels = metric.take_label();`

meqif: That doesn't compile with the default features:

```
$ cargo build
   Compiling prometheus v0.6.0 (/Users/meqif/rust-prometheus)
error[E0599]: no method named `append` found for type `protobuf::RepeatedField<proto::LabelPair>` in the current scope
   --> src/registry.rs:209:32
    |
209 |                         labels.append(&mut pairs);
    |                                ^^^^^^

```

breeswish: Oh. I got you.

meqif:
Thanks for the review!

> Would you like to also modify travis file to make CI test for compilation without protobuf in this PR?

I don't quite know how the matrix configuration for Travis should be configured, but I can give it a shot.

meqif:
Cool, I think I got it right, although it doesn't seem to apply to all Rust channels (but that's probably because the channels are listed under `matrix` and not under `rust`, from what I gather in Travis' documentation). The tests did reveal another instance of `LabelPair::new()` that should have been replaced with `LabelPair::default()`, which I've fixed.

@breeswish Do you mind taking another look at the changes?

breeswish:
Awesome! Thanks a lot!

PR #251, opened 2019-06-20
Provide immutable interface for local counter

breeswish:
When local metrics are used in thread local environments, it would be better to provide immutable interfaces, so that there is no need to wrap local metrics with a `RefCell`.

`LocalHistogram` is already immutable, so this PR makes `Local(Int)Counter` immutable.


lucab:
Gentle reminder to push a git-tag too.

PR #257, opened 2019-07-27
histogram: augment timers

lucab:
This augments histogram timers (both sync and unsync) with several
new features:
 * timers can now return their duration on stop, in order for
   consumers to use that as the input for further logic.
 * timers can now be stopped and discarded, in order to (optionally)
   not record their duration to the attached histogram.

All of the above come with corresponding tests.

### DIFF
--- a/src/histogram.rs
+++ b/src/histogram.rs
@@ -304,48 +304,80 @@ mod coarse {
     }
 }
 
-/// A struct represents an event being timed. When the timer goes out of
-/// scope, the duration will be observed, or call `observe_duration` to manually
-/// observe.
+/// Timer to measure and record the duration of an event.
 ///
-/// NOTICE: A timer can be observed only once (automatically or manually).
+/// This timer can be stopped and observed at most once, either automatically (when it
+/// goes out of scope) or manually.
+/// Alternatively, it can be manually stopped and discarded in order to not record its value.
 #[must_use = "Timer should be kept in a variable otherwise it cannot observe duration"]
 pub struct HistogramTimer {
+    /// An histogram for automatic recording of observations.
     histogram: Histogram,
+    /// Whether the timer has already been observed once.
+    observed: bool,
+    /// Starting instant for the timer.
     start: Instant,
 }
 
 impl HistogramTimer {
-    fn new(histogram: Histogram) -> HistogramTimer {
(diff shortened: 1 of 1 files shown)


lucab:
/cc @overvenus @breeswish for review.

My usecase for this is a slightly complex scenario (no code to link yet, sorry) where I want to have a standalone timer and manually dispatch its value to different histograms/labels, depending on several other conditions.

lucab:
This is temporarily based on https://github.com/pingcap/rust-prometheus/pull/259 as plain master is otherwise broken.

breeswish:
Thanks for your contribution! After this modification, is the detached `HistogramTimer` exactly the same as a [coarsetime](https://github.com/jedisct1/rust-coarsetime) timer?

lucab:
Rebased on latest master.

@breeswish indeed the "standalone" part of this PR makes it similar to any generic timer. So yes, `HistogramTimer` now looks like a bit like `coarsetime`, except I didn't change the coarseness so it is likely still unoptimized and more precise.
At this point I don't have performance/precision concerns, so I'm fine with whatever defaults this library decides to use. The reason for this PR is just to make time measurements orthogonal to recordings. I'm already using the timer from this library everywhere else in my codebase and would like to stick to them also for the cases that are more complex than what implicit recording allows.

breeswish:
Thanks for the reply. I think introducing standalone elapsed time measurement interface for HistogramTimer is not a good design currently. As its name indicates, it is a timer for histogram, but obviously not a timer for general time measurement. These newly added interface may lead to confusion for normal users. For your use case, I recommend to use coarsetime directly, whose value can then be fed to Histogram::observe().

lucab:
@breeswish ack, I see the rationale behind that, I will change my consumer to use a dedicated library for timers.

Regarding the other two points in this PR (discarding a running timer and returning the observed value), do you think those are still useful to have? Should I amend this PR and just drop the standalone-timer part?

breeswish:
@lucab Yes! I think it is a good idea. You may remove the `is_standalone` and `start_standalone`. Other newly introduced functions seems to be very clear!

lucab:
@breeswish I've rebased and dropped all the "standalone" bits, PTAL.

PR #261, opened 2019-08-01
counter: add support for runtime resets (zeroing)

lucab:
Counters can only increase or be reset to zero on restart.
This adds support for soft resets at runtime, by implementing and
exposing reset-to-zero methods.

Ref: https://prometheus.io/docs/concepts/metric_types/#counter

### DIFF
--- a/src/counter.rs
+++ b/src/counter.rs
@@ -86,6 +86,12 @@ impl<P: Atomic> GenericCounter<P> {
         self.v.get()
     }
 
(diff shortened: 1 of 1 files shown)


lucab:
/cc @overvenus @breeswish @siddontang  for review.

My immediate usecase for this is related to testing. However I think there are other scenarios (e.g. dynamic configuration reloading) where it could be useful.

lucab:
@siddontang I think this one is ready to merge.

PR #268, opened 2019-08-22
Update Syn

nrc:
Also fixes a compile issue with the `nightly` feature and updates procinfo to use the same fork as TiKV.

PTAL @breeswish  @Hoverbear

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus"
-version = "0.7.0"
+version = "0.8.0"
 license = "Apache-2.0"
 keywords = ["prometheus", "metrics"]
 authors = ["overvenus@gmail.com", "siddontang@gmail.com", "vistaswx@gmail.com"]
--- a/static-metric/Cargo.toml
+++ b/static-metric/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-static-metric"
-version = "0.2.0"
+version = "0.3.0"
 license = "Apache-2.0"
 authors = ["me@breeswish.org"]
 description = "Static metric helper utilities for rust-prometheus."
@@ -17,13 +17,13 @@ include = [
 proc-macro = true
 
 [dependencies]
-syn = { version = "0.13", features = ["full", "extra-traits"] }
-proc-macro2 = "0.3"
-quote = "0.5"
+syn = { version = "0.15", features = ["full", "extra-traits"] }
+proc-macro2 = "0.4"
+quote = "0.6"
 lazy_static = "1.0"
 
 [dev-dependencies]
-prometheus = "0.7"
+prometheus = { path = "../" }
 coarsetime = "0.1"
 
 [features]
--- a/static-metric/src/parser.rs
+++ b/static-metric/src/parser.rs
@@ -14,9 +14,9 @@
 use std::collections::HashMap;
 
 use proc_macro2::Span;
-use syn::buffer::Cursor;
+use syn::parse::{Parse, ParseStream};
 use syn::punctuated::Punctuated;
-use syn::synom::{PResult, Synom};
+use syn::token::*;
 use syn::*;
 
 /// Matches `label_enum` keyword.
@@ -24,18 +24,13 @@ struct LabelEnum {
     pub span: Span,
 }
 
-impl Synom for LabelEnum {
-    fn parse(tokens: Cursor) -> PResult<Self> {
-        if let Some((term, rest)) = tokens.term() {
-            if term.as_str() == "label_enum" {
-                return Ok((LabelEnum { span: term.span() }, rest));
-            }
+impl Parse for LabelEnum {
+    fn parse(input: ParseStream) -> Result<Self> {
+        let ident: Ident = input.parse()?;
+        if &ident == "label_enum" {
+            return Ok(LabelEnum { span: ident.span() });
         }
-        parse_error()
-    }
-
-    fn description() -> Option<&'static str> {
-        Some("label_enum")
+        Err(input.error("Expected `label_enum`"))
     }
 }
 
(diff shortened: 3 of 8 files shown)


Cargo.toml
```diff
@@ -35,7 +35,7 @@ spin = "0.5"
 reqwest = { version = "0.9.5", optional = true }
 
 [target.'cfg(target_os = "linux")'.dependencies]
-procinfo = { version = "0.4", optional = true }
+procinfo = { git = "https://github.com/tikv/procinfo-rs", optional = true }
```
breeswish: maybe not appropriate to use pingcap's own procinfo?

static-metric/Cargo.toml
```diff
@@ -17,9 +17,9 @@ include = [
 proc-macro = true
 
 [dependencies]
-syn = { version = "0.13", features = ["full", "extra-traits"] }
-proc-macro2 = "0.3"
-quote = "0.5"
+syn = { version = "0.15", features = ["full", "extra-traits"] }
```
breeswish: syn has version 1.0.3: https://docs.rs/syn/1.0.3/syn/

static-metric/Cargo.toml
```diff
@@ -17,9 +17,9 @@ include = [
...
+proc-macro2 = "0.4"
+quote = "0.6"
```
breeswish: quote has version 1.0.2: https://docs.rs/quote/1.0.2/quote/

static-metric/Cargo.toml
```diff
@@ -17,9 +17,9 @@ include = [
 proc-macro = true
 
 [dependencies]
-syn = { version = "0.13", features = ["full", "extra-traits"] }
-proc-macro2 = "0.3"
-quote = "0.5"
+syn = { version = "0.15", features = ["full", "extra-traits"] }
+proc-macro2 = "0.4"
```
breeswish: proc-macro2 has version 1.0.1: https://docs.rs/proc-macro2/1.0.1/proc_macro2/

Cargo.toml
nrc: Why not? We use it in TiKV

lucab: But `tikv/procinfo-rs` is not published as a crate, right? Which would make this crate also not publishable anymore.
I do personally have projects (already on crates.io) which need to consume this from the registry.

Have you maybe considered switching to https://docs.rs/procfs/0.5.3/procfs/ or checking with @danburkert if they are open to co-maintainers? (does PingCAP want to maintain that crate?)

static-metric/Cargo.toml
kennytm: IMO we should update TiKV's `syn` to 1.0 too.

Cargo.toml
nrc: I can change this back if we think it is really important, the key thing is that we move off 0.3 which the version of prometheus in TiKV is currently using

static-metric/Cargo.toml
nrc: The issue is not what version TiKV is using, but what version it's deps are using (Serde, Prost, slog, structopt, and more). We *should* send PRs to update them all, but I don't want to take that on right now when there is low-hanging fruit in just updating Prometheus to use a more recent version. Let's not let perfect be the enemy of good here.

nrc: ping @breeswish

breeswish: Ok

breeswish: But what's the purpose of updating these versions?

nrc:
ping @Hoverbear @breeswish changed procinfo to version 4 so we can still publish the crate. PTAL.

static-metric/src/parser.rs
```diff
@@ -14,28 +14,23 @@
...
-impl Synom for LabelEnum {
-    fn parse(tokens: Cursor) -> PResult<Self> {
-        if let Some((term, rest)) = tokens.term() {
-            if term.as_str() == "label_enum" {
-                return Ok((LabelEnum { span: term.span() }, rest));
-            }
+impl Parse for LabelEnum {
+    fn parse(input: ParseStream) -> Result<Self> {
+        let ident: Ident = input.parse()?;
+        if &ident.to_string() == "label_enum" {
```
Hoverbear: Is the `.to_string()` here an unnecessary copy?

nrc: It is a copy, I'm not sure if it is unnecessary - I don't think there is a better way of getting a string repr of an `Ident`, and turning a string into an ident also requires a copy. It also doesn't matter - the ident name will always be short and this code only executes at compile time and is not hot.

breeswish:
LGTM. I prefer to add testing script of the static-metric to the travis as well (e.g. building the library as well as its examples) to make sure that everything is ok.

static-metric/src/parser.rs
kennytm: @nrc `syn 0.15.44` has [this impl](https://docs.rs/syn/0.15.44/syn/struct.Ident.html#impl-PartialEq%3CT%3E):

```rust
impl<T> PartialEq<T> for Ident where
    T: AsRef<str> + ?Sized, 
```

so `ident == "label_enum"` (add or minus a few `&` and `*`) should just work.

nrc:
ping @Hoverbear @breeswish PTAL

breeswish:
How about the https://github.com/pingcap/rust-prometheus/pull/268#issuecomment-528170310 ?

nrc:
> How about the #268 (comment) ?

@breeswish  Its is addressed in https://github.com/pingcap/rust-prometheus/pull/268/files#diff-354f30a63fb0907d4ad57269548329e3, I think

nrc:
@kennytm PTAL (I've addressed @Hoverbear's and your comments)

PR #269, opened 2019-08-28
Fix bug in global label handling.

brndnmtthws:
There was a bug in the label handling where every metric after the first
did not have labels applied.

We were calling Vec::append() with a mutable vec as the argument, and
doing so causes the elements to change ownership from the source to the
target vec.

This change fixes the problem by cloning the label vec and introducing a
test.


lucab:
Sorry for not catching this mistake in my PR :(
@breeswish @overvenus can this be merged now?

PR #270, opened 2019-08-28
Fix compile on nightly.

brndnmtthws:
LocalHistogramTimer::new_coarse() should take a local histogram as an
arg.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,8 @@ matrix:
   fast_finish: true
   include:
   - rust: nightly
+  - rust: nightly
+    env: FEATURES="nightly protobuf push process"
   - rust: beta
   - rust: stable
     install:
--- a/src/histogram.rs
+++ b/src/histogram.rs
@@ -652,7 +652,7 @@ impl LocalHistogramTimer {
     }
 
     #[cfg(feature = "nightly")]
-    fn new_coarse(histogram: Histogram) -> Self {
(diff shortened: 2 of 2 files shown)


breeswish:
Thanks a lot. Would you also like to modify our Travis CI so that nightly can be included and checked in CI?

brndnmtthws:
> Thanks a lot. Would you also like to modify our Travis CI so that nightly can be included and checked in CI?

Done!

brndnmtthws:
Woops, not so fast, that doesn't work. One moment.

brndnmtthws:
I think the travis build needs a kick.

breeswish:
Hmm.. Travis status is simply missing.. OMG

.travis.yml
```diff
@@ -37,6 +37,10 @@ matrix:
   - rust: nightly
 
 script:
+  - if [ $TRAVIS_RUST_VERSION = 'nightly' ]; then
```
breeswish: Maybe we can use the include feature: https://docs.travis-ci.com/user/build-matrix/#excluding-jobs

e.g.

```yaml
include:
- rust: nightly
- rust: nightly
  env: FEATURES="nightly protobuf push process"
- rust: stable
```

brndnmtthws: Ah, yeah, that looks cleaner, didn't know about that.

brndnmtthws:
Rebased on master, and updated to use the travis matrix feature instead. Seems travis is back from its hiatus as well.

PR #271, opened 2019-08-28
Wrap timespec with struct to implement fmt::Debug.

brndnmtthws:


breeswish:
Thanks a lot!

Looks like we need to cover different features in CI @overvenus

PR #273, opened 2019-09-04
Add support to set timestamp in ms of the metrics

lfdominguez:
There some situations where we need to put the exact timestamp where the data is collected. Example when your exporter get information from other source, like an APIRest that has the real timestamp of the data.


Hoverbear:
This is related to supporting the `timestamp()` function? https://prometheus.io/docs/prometheus/latest/querying/functions/#timestamp

nagisa:
This is potentially racy, isn’t it? E.g. 

1. Thread A sets the value;
2. Thread B sets the value;
3. Thread B sets the timestamp;
4. Thread A sets the timestamp.

causing the final metric value to have value from Thread B but timestamp for an earlier value from thread A.

Setting a value and the timestamp should be done with a single method call.

It is also not clear to me how this timestamp is presented in the gathered data.

lfdominguez:
> This is related to supporting the `timestamp()` function? https://prometheus.io/docs/prometheus/latest/querying/functions/#timestamp

Noup

mxinden:
I am closing here given that the discussion has stalled and the fact that this would introduce a race-condition.

@lfdominguez in case you would like to further discuss this, please reopen the pull request or open a new issue.

PR #274, opened 2019-09-05
Update spin dependency to 0.5.2

ekse:
A vulnerability was fixed in spin and is reported by cargo-audit. Although exploitation sounds unlikely, the warning shows up transitively for packages that depend on rust-prometheus, so updating it would be nice.

   https://rustsec.org/advisories/RUSTSEC-2019-0013.html


breeswish:
Thanks a lot!

PR #275, opened 2019-09-15
export local inner macros for rust2018 use

qmx:
when using this crate on a rust 2018 project, some macros like `register_counter_vec` will fail with a very confusing error - that's because you need to bring the helper `__register_counter_vec` into scope.

with this change this rust 2018 style code works just fine:

```
fn main() {
    prometheus::register_counter!("test_macro_2", "help");
    prometheus::register_counter_vec!("errorz", "errors", &["error"]).unwrap();
    prometheus::register_gauge!("test_macro_gauge_2", "help");
    prometheus::register_gauge_vec!("test_macro_gauge_vec_3", "help", &["a", "b"]);
    prometheus::register_histogram!("test_macro_histogram_4", "help", vec![1.0, 2.0]);
    prometheus::register_histogram_vec!(
        "test_macro_histogram_4",
        "help",
        &["a", "b"],
        vec![1.0, 2.0]
    );
    prometheus::register_int_counter!("meh", "foo");
    prometheus::register_int_counter_vec!("errorz", "errors", &["error"]).unwrap();
    prometheus::register_int_gauge!("test_macro_gauge_2", "help");
    prometheus::register_int_gauge_vec!("test_macro_gauge_vec_4", "help", &["a", "b"]);
}
```

### DIFF
--- a/examples/example_edition_2018.rs
+++ b/examples/example_edition_2018.rs
@@ -1,0 +1,22 @@
+use prometheus::register_counter;
+
+/// small example that uses rust 2018 style macro imports
+
+#[allow(unused)]
+fn main() {
+    register_counter!("test_macro_3", "help");
+    prometheus::register_counter_vec!("errorz", "errors", &["error"]).unwrap();
+    prometheus::register_gauge!("test_macro_gauge_2", "help");
+    prometheus::register_gauge_vec!("test_macro_gauge_vec_3", "help", &["a", "b"]);
+    prometheus::register_histogram!("test_macro_histogram_4", "help", vec![1.0, 2.0]);
+    prometheus::register_histogram_vec!(
+        "test_macro_histogram_4",
+        "help",
+        &["a", "b"],
+        vec![1.0, 2.0]
+    );
+    prometheus::register_int_counter!("meh", "foo");
+    prometheus::register_int_counter_vec!("errorz", "errors", &["error"]).unwrap();
+    prometheus::register_int_gauge!("test_macro_gauge_2", "help");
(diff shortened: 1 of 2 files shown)


qmx:
@breeswish sure - can you point me to where you want a test like this inside the codebase? (I didn't dive into it because time 😛 )

breeswish:
@qmx Maybe you can add an example called edition_2018, where you use `use prometheus::xxx` style to import the macro instead of the traditional `#[macro_use]`. Then you can modify the travis to cover examples so that compiling failures will result in CI failure.

qmx:
@breeswish how does that look now?

.travis.yml
```diff
@@ -40,6 +40,7 @@ matrix:
 
 script:
   - cargo test --no-default-features --features="$FEATURES"
```
breeswish: maybe you can consider: `cargo test --all-targets` then examples will be covered.

qmx: we actually don't need `--all-targets` at all - the default test profile compiles the example files too during it's build step.

tested locally by throwing invalid syntax inside the file, and watching it break.
pushed cleaned up commits

breeswish: Alright :)

breeswish:
Thanks a lot!

p-shahi:
@breeswish I'm running into this issue on version 0.7.0 on stable 1.39.0 build.
When I use the `register_int_gauge! ` macro,  I get the following compile error
> error: cannot find macro `__register_gauge` in this scope

and have to import `__register_gauge` to resolve it.
I believe this pull request fixes that issue.
Do you know what the release schedule for this crate is? Is it possible to update crates.io with this fix in a minor release?

PR #284, opened 2019-10-10
histogram: add start_timer_precision api

wayslog:
this PR make it possible to change the precise of HistogramTimer.

example:

```rust
let opts = HistogramOpts::new("test1", "test help")
            .const_label("a", "1")
            .const_label("b", "2");
let histogram = Histogram::with_opts(opts).unwrap();
let timer = histogram.start_timer_precise(TimerPrecise::Millis);
thread::sleep(Duration::from_millis(100));
timer.observe_duration();

let mut mfs = histogram.collect();
assert_eq!(mfs.len(), 1);
let mf = mfs.pop().unwrap();
let m = mf.get_metric().get(0).unwrap();
let proto_histogram = m.get_histogram();
assert!(proto_histogram.get_sample_sum() > 999.0;
```

### DIFF
--- a/src/histogram.rs
+++ b/src/histogram.rs
@@ -325,6 +325,34 @@ mod coarse {
     }
 }
 
+/// TimerPrecise is the precise for HistogramTimer
+///
+/// Default start_timer has a precise of seconds, you can change it by using
+/// `start_timer_precise`.
+///
+#[derive(Debug, Clone, Copy)]
+pub enum TimerPrecise {
+    /// Second precise. means that a Duration::from_secs(1) can be represent as 1.0f64
+    Second,
+    /// Millis is millisecond, means that Duration::from_millis(199) can be represent as 100.0f64.
+    Millis,
+    /// Micros is microsecond, means that Duration::from_millis(199) can be represent as 199000.0f64.
+    Micros,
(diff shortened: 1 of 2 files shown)


src/histogram.rs
```diff
@@ -335,25 +363,29 @@ mod coarse {
 pub struct HistogramTimer {
     /// An histogram for automatic recording of observations.
     histogram: Histogram,
+    /// Precise of this timer and default is [Second](TimerPrecise::Second)
+    precise: TimerPrecise,
```
kennytm: As a noun it should be called "precision".

breeswish:
Thanks for the PR! Do you think it is better to call it something like "observe_unit" instead of "precision"? Because the timer has a *fixed* precision, determined by the precision of Instant, which is a nanosecond precision.

mxinden:
I don't see how adding precision to `Histogram` timers is of much use. E.g. with `start_coarse_timer` a user can already control the performance overhead introduced. For users wanting to expose anything other than base units (which is not recommended for compatibility reasons) they can use the `observe` code path.

I am closing here. @wayslog feel free to reopen the pull request or create a new one in case you would like to further discuss this.

PR #285, opened 2019-10-11
Use thiserror instead of quick-error

koushiro:

- Use failure instead of quick-error
- Fix some typos
- Reorder some import

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -20,8 +20,13 @@ pub enum Error {
     #[error("Duplicate metrics collector registration attempted")]
     AlreadyReg,
     /// The label cardinality was inconsistent.
-    #[error("Inconsistent label cardinality, expect {0} label values, but got {1}")]
-    InconsistentCardinality(usize, usize),
+    #[error("Inconsistent label cardinality, expect {expect} label values, but got {got}")]
+    InconsistentCardinality {
+        /// The expected number of labels.
+        expect: usize,
+        /// The actual number of labels.
+        got: usize,
+    },
     /// An error message which is only a string.
     #[error("Error: {0}")]
     Msg(String),
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -11,76 +11,44 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::io::Error as IoError;
-use std::result;
-
-#[cfg(feature = "protobuf")]
-use protobuf::error::ProtobufError;
-
-#[cfg(feature = "protobuf")]
-quick_error! {
-    /// The error types for prometheus.
-    #[derive(Debug)]
-    pub enum Error {
-        /// A duplicate metric collector has already been registered.
-        AlreadyReg {
-            description("duplicate metrics collector registration attempted")
-        }
-        /// The label cardinality was inconsistent.
-        InconsistentCardinality(expect: usize, got: usize) {
-            description("inconsistent label cardinality")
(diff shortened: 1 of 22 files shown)


kennytm:
PTAL @breeswish you suggested `failure` in https://github.com/pingcap/rust-prometheus/issues/218#issuecomment-458882638. 

However, `failure` is pretty much a failure now (its author has released `fehler` as the latest experiment), I question whether this is still a correct replacement. 

(See also https://www.reddit.com/r/rust/comments/dfs1zk/2019_q4_error_patterns_snafu_vs_errderive_anyhow/ for the latest discussion.)

breeswish:
I agree with @kennytm that failure is not the best choice now. We have many new choices like [fehler](https://github.com/withoutboats/), [thiserror](https://github.com/dtolnay/thiserror) and [snafu](https://github.com/shepmaster/snafu). I guess thiserror should be used instead as an error used in library. What do you think? @kennytm

koushiro:
I tried the [thiserror](https://github.com/dtolnay/thiserror) crate, which is awesome. I will use it instead of `failure` if @kennytm agrees with that.

kennytm:
SGTM. Comparing the three, `thiserror` is the lightest, and also the most stable (in terms of semver).

`fehler` is more like an experiment (to fit exceptions into Rust). It isn't a good idea for us to depend on an unstable design pattern. 

`snafu` has quite a lot of features, and thus 3× the crate size of the other two. Like `failure` it introduced a `.context()` method and backtrace. These would be useful for TiKV, though here in `rust-prometheus` we didn't use these.

koushiro:
@breeswish @kennytm PTAL :)

src/errors.rs
```diff
@@ -11,76 +11,28 @@
...
-    }
+/// The error types for prometheus.
+#[derive(Debug, Error)]
+pub enum Error {
+    /// A duplicate metric collector has already been registered.
+    #[error("Duplicate metrics collector registration attempted")]
+    AlreadyReg,
+    /// The label cardinality was inconsistent.
+    #[error("Inconsistent label cardinality, expect {0} label values, but got {1}")]
+    InconsistentCardinality(usize, usize),
```
breeswish: Maybe we should name it

src/errors.rs
```diff
@@ -11,76 +11,28 @@
...
+    /// An error message which is only a string.
+    #[error("Error: {0}")]
+    Msg(String),
```
breeswish: In future PRs we should clean up this kind of errors.

src/errors.rs
```diff
@@ -11,76 +11,28 @@
...
+    Io(#[from] std::io::Error),
+    /// An error containing a [`protobuf::Error`].
+    #[cfg(feature = "protobuf")]
+    #[error("Protobuf error: {0}")]
+    Protobuf(#[from] protobuf::error::ProtobufError),
 }
 
 /// A specialized Result type for prometheus.
-pub type Result<T> = result::Result<T, Error>;
+pub type Result<T> = std::result::Result<T, Error>;
```
breeswish: The size could be a bit large even if it is `Result<()>`..

koushiro: Yes, do you have any good idea? :(

breeswish: Nop... @kennytm Ideas?

kennytm: You could `Box` some components. But you need to define what is meant by "Large" 🙃

breeswish:
@overvenus PTAL

Issue #287, opened 2019-10-14
unexpected status code 200 OK when using push_metrics


hvariant:
I'm getting `Msg("unexpected status code 200 OK while pushing to ...")` when pushing to prometheus pushgateway docker [v0.10.0](https://hub.docker.com/layers/prom/pushgateway/latest/images/sha256-5e252ed423c397fcab03c718714464a0a8f8a0b51e075f38a38519a3eac8f171).

**To Reproduce**
Run the push example to send gauge metric to a pushgateway running in a docker container.

**Actual behavior**
Crash at `push_metrics(...).unwrap` because it's returning `Err(Msg("unexpected status code 200 OK while pushing to ..."))` instead of `Ok(_)`.

**Expected behavior**
No crash happens. 200 Response should be an acceptable response since the push was successful and I can query the metrics in prometheus.

**System information**
* CPU architecture: x86_64
* Distribution and kernel version: 5.3.5-arch1-1-ARCH
* SELinux on?: no
* Any other system details we should know?: pushgateway is running in [docker](https://hub.docker.com/layers/prom/pushgateway/latest/images/sha256-5e252ed423c397fcab03c718714464a0a8f8a0b51e075f38a38519a3eac8f171).

**Additional Info**
Not sure if this is relevant but I'm not using any authentication.

hvariant:
from pushgateway [README](https://github.com/prometheus/pushgateway/blob/master/README.md):

```
The response code upon success is either 200 or 400. A 200 response implies a successful push, either replacing an existing group of metrics or creating a new one. A 400 response can happen if the request is malformed or if the pushed metrics are inconsistent with metrics pushed to other groups or collide with metrics of the Pushgateway itself. An explanation is returned in the body of the response and logged on error level.
```

orfest:
Same with pushgateway v1.0

hvariant:
> Same with pushgateway v1.0

Yeah, the official go client has a comment about this breaking change: https://github.com/prometheus/client_golang/blob/master/prometheus/push/push.go#L270

hvariant:
I should probably close this issue since my change is merged

sebastianjonasson:
Is this patch gonna be released anytime soon?

rustonaut:
Run into this just now, good to know it's fixed (but not yet pushed to cargo as far as I can tell).

pankhudinigam:
Any timelines on this fix being available to users?

mxinden:
[...]

PR #288, opened 2019-10-14
accept 200 response code when pushing to pushgateway

hvariant:
should I remove `StatusCode::ACCEPTED => Ok(())`?


hvariant:
this fixes #287

hvariant:
also found in the go client: https://github.com/prometheus/client_golang/blob/master/prometheus/push/push.go#L270

PR #289, opened 2019-10-15
Update syn, quote, proc-macro2 to v1.0

koushiro:

I think it's time to update `syn`, `quote` and `proc-macro2` to v1.0.
`syn`, `quote` and `proc-macro2` in the upstream dependency of `rust-prometheus` have been upgraded to version 1.0.


breeswish:
Thanks!

PR #290, opened 2019-10-26
Switch from procinfo to procfs

### LINKED ISSUE #272: Consider switching from procinfo to procfs
The procinfo crate seems somewhat inactive, and there's an alternative called [procfs](https://github.com/eminence/procfs) which appears more well maintained.

It would be great to be able to capture additional process metrics for observability.

Additionally, it would be nice to mix OS level metrics in with process metrics. For example, I want to know how the process is affecting CPU pressure on the host system.

eminence:
This commit doesn't add any new features, but allows future commits to
use features in procfs that are missing from procinfo.  Also hopefully rust-prometheus won't have to maintain as much code to deal with process info.

The idea for this change came from @brndnmtthws, who wanted to use some features in procfs that are missing from procinfo.  I haven't added these features yet, but if this PR is accepted, I can help with that work too.   Disclaimer: I am the author of the procfs crate.

Closes #272

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ features = ["nightly"]
 default = ["protobuf"]
 nightly = ["libc"]
 push = ["reqwest", "libc", "protobuf"]
-process = ["libc", "procinfo"]
+process = ["libc", "procfs"]
 gen = ["protobuf-codegen-pure"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


eminence:
The conversation in #272 is leaning toward a new crate for host metrics.  But this has got me thinking that a new crate might also be a good idea for process metrics using procfs:

* You don't have to do anything
* Users get to choose to use the built-in process collector, or an external crate if they want additional features
* You could still roll in a 3rd party crate into rust-prometheus in the future if you wanted to.

Given that the current code is working fine for you, perhaps the best thing to do is decline this pull request (unless you are looking forward to dropping some of the code I removed)

PR #292, opened 2019-11-07
add unsigned integer gauge

hskang9:


breeswish:
Thanks for the PR! Why would you want a UIntGauge?

hskang9:
The thing is that I was working on putting prometheus in a blockchain framework [Substrate]() with @nodebreaker-carlsagan, and found out there is only IntGauge option for sending `u64` value. In my perspective, The maximum number of i64 is 2^63 - 1:    0111111111111111111111111111111111111111111111111111111111111111 excluding parity bit for negative number representation where as the maximum number is 
2^64 - 1:    1111111111111111111111111111111111111111111111111111111111111111. 
In a blockchain where the block number(which is `u64`) that can increase endlessly, I thought `UIntGauge` should be made to cover the gap between 1 binary digit.

breeswish:
@hskang9 For this case, I suggest to use Gauge directly, since it IntGauge is only an optimization in rust-prometheus and is opaque to Prometheus. Prometheus always treat them as f64 instead.

hskang9:
I see. Okay closing PR for now.

PR #294, opened 2019-11-13
use Cow where necessary

fredchenbj:
Signed-off-by: fredchenbj <cfworking@163.com>

for https://github.com/tikv/rust-prometheus/issues/282 and https://github.com/tikv/tikv/issues/5705, try to use Cow<'static, str> to improve the memory usage.

`make all` passed.

### DIFF
--- a/src/counter.rs
+++ b/src/counter.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::HashMap;
 use std::marker::PhantomData;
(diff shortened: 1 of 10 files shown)


fredchenbj:
PTAL. @breeswish

breeswish:
Oh, sorry, I missed this PR for so long!

Your PR have some issues in that your change actually does not always improve performance. For example, as long as you have written "some_cow_string.to_string()", then there is the copy cost from `str` to `String` if `COW` stores `str` and it becomes as slow as calling `str.to_string()` directly.

mxinden:
I agree with @breeswish that this is not a performance improvement as those values need to be converted to owned values eventually anyways.

I am linking https://github.com/tikv/rust-prometheus/pull/308#pullrequestreview-434735225 here as I find it a more promising approach.

@fredchenbj I am closing here. Feel free to reopen the pull request in case I am missing something or in case you would like to further discuss this change.

PR #298, opened 2019-12-15
Add warp example

marcelbuesing:
This adds an example that uses warp for serving the metric data. It's based on the hyper example.
Currently still waiting for this PR https://github.com/seanmonstar/warp/pull/336.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ procfs = { version = "0.7", optional = true, default-features = false }
 getopts = "0.2"
 hyper = "0.13"
 tokio = { version = "0.2", features = ["macros"] }
+warp = "0.2"
 
 [build-dependencies]
 protobuf-codegen-pure = { version = "2.0", optional = true }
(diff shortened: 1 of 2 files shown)


breeswish:
Thanks! Looks like seanmonstar/warp#336 is merged. Would you like to update your PR?

mxinden:
@marcelbuesing @breeswish I don't feel like another example with an http library that is based on the library used in the already existing example is of much help to users. Is there something in particular you would like to stress that is not already explained with https://github.com/tikv/rust-prometheus/blob/master/examples/example_hyper.rs?

lucab:
Additional concern, the more dependencies we add to Cargo manifest (including dev-deps) the bigger it becomes a burden for packagers and distributors to handle. I'd rather go in the other direction, where we try to minimize the set of examples with external dependecies, or move them to a dedicated out-of-tree examples crate.

marcelbuesing:
Sure I can certainly understand the reasoning ! I assume people can still find this PR when searching for this in particular.

PR #302, opened 2019-12-17
Correct intra-doc links

quodlibetor:
Wherever possible, use the default linking style of looking names up in the
enclosing namespace without adding an explicit link, but sometimes we need to
add the link target via parens, in those cases correct from `::Name` to
`crate::Name`.

### DIFF
--- a/src/atomic64.rs
+++ b/src/atomic64.rs
@@ -7,7 +7,7 @@ use std::ops::*;
 use std::sync::atomic::{AtomicI64 as StdAtomicI64, AtomicU64 as StdAtomicU64, Ordering};
 
 /// An interface for numbers. Used to generically model float metrics and integer metrics, i.e.
-/// [`Counter`](::Counter) and [`IntCounter`](::IntCounter).
+/// [`Counter`](crate::Counter) and [`IntCounter`](crate::Counter).
 pub trait Number:
     Sized + AddAssign + SubAssign + PartialOrd + PartialEq + Copy + Send + Sync
 {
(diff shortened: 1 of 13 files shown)


breeswish:
Thanks for the contribution! There are still some remaining unresolved links. Would you like to fix it?

```
warning: `[::Counter]` cannot be resolved, ignoring it...
   --> src/counter.rs:230:52
    |
230 |     /// Flush the local metrics to the [`Counter`](::Counter).
    |                                                    ^^^^^^^^^ cannot be resolved, ignoring
    |
    = note: `#[warn(intra_doc_link_resolution_failure)]` on by default
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[::CounterVec]` cannot be resolved, ignoring it...
   --> src/counter.rs:300:55
    |
300 |     /// Flush the local metrics to the [`CounterVec`](::CounterVec) metric.
    |                                                       ^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[::Histogram]` cannot be resolved, ignoring it...
   --> src/histogram.rs:881:54
    |
881 |     /// Flush the local metrics to the [`Histogram`](::Histogram) metric.
    |                                                      ^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[::HistogramVec]` cannot be resolved, ignoring it...
   --> src/histogram.rs:933:57
    |
933 |     /// Flush the local metrics to the [`HistogramVec`](::HistogramVec) metric.
    |                                                         ^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`
```

quodlibetor:
Yes definitely! I thought I checked for everything.

quodlibetor:
I believe that all links are correct now.

PR #308, opened 2020-02-24
Use AsRef<str> for label values to make it easier to use with String …

AlexanderThaller:
Sometimes labels get constructed from existing values. Those values can
be of the type String instead of &str. Using AsRef<str> instead of &str
in the parameters where labels are passed allows both the usage of &str
and String.

I'm not totally sure if this makes sense. I had places where a empty &[] is passed that needed some type annotation and I'm not sure if constructing a Vec is the best solution there.

Besides that benchmarking dind't show much of a difference performance wise, at least on my machine.

```
» cargo benchcmp with_str_slice with_as_ref_str_slice
 name                                              with_str_slice ns/iter  with_as_ref_str_slice ns/iter  diff ns/iter   diff %  speedup
 bench_atomic_f64                                  8                       8                                         0    0.00%   x 1.00
 bench_atomic_i64                                  3                       3                                         0    0.00%   x 1.00
 bench_counter_no_labels                           8                       8                                         0    0.00%   x 1.00
 bench_counter_no_labels_concurrent_nop            8                       8                                         0    0.00%   x 1.00
 bench_counter_no_labels_concurrent_write          302                     298                                      -4   -1.32%   x 1.01
 bench_counter_with_label_values                   37                      39                                        2    5.41%   x 0.95
 bench_counter_with_label_values_concurrent_write  418                     404                                     -14   -3.35%   x 1.03
 bench_counter_with_mapped_labels                  161                     156                                      -5   -3.11%   x 1.03
 bench_counter_with_prepared_mapped_labels         98                      82                                      -16  -16.33%   x 1.20
 bench_gauge_no_labels                             8                       8                                         0    0.00%   x 1.00
 bench_gauge_with_label_values                     39                      38                                       -1   -2.56%   x 1.03
 bench_histogram_local                             5                       5                                         0    0.00%   x 1.00
 bench_histogram_no_labels                         17                      16                                       -1   -5.88%   x 1.06
 bench_histogram_timer                             67                      67                                        0    0.00%   x 1.00
 bench_histogram_with_label_values                 51                      47                                       -4   -7.84%   x 1.09
 bench_int_counter_no_labels                       3                       3                                         0    0.00%   x 1.00
 bench_int_counter_no_labels_concurrent_write      66                      76                                       10   15.15%   x 0.87
 bench_int_gauge_no_labels                         3                       3                                         0    0.00%   x 1.00
 bench_local_histogram_timer                       98                      99                                        1    1.02%   x 0.99
```


breeswish:
Thanks for the contribution!

- You can use `impl AsRef<str>` in most places.
- You need to amend your commit to include the sign-off in order to pass the DCO check.

The rest LGTM!

AlexanderThaller:
Thank you for the review!
I updated the pullrequest to use impl AsRef<str> everywhere.
I also changed the empty_labels definition to create a empty slice. Somehow that feels cleaner to do than creating a Vec.

AlexanderThaller:
Merged the documentation changes from upstream master should work now again.

AlexanderThaller:
I can try to cleanup the history if that merge commit is too ugly :D

AlexanderThaller:
@breeswish Can this be merged maybe? :D

AlexanderThaller:
That makes sense to me. Im gonna try to implement it with Into<String> to communicate the allocation better. After that I'm gonna try to rebase from master again.

Sry for the long time between replies.

AlexanderThaller:
I did most of the changes to Into<String>. I had to sometimes further restrict the type to allow as_bytes() at one place. Now I'm running into the following error which im not sure how to fix...

```
error[E0508]: cannot move out of type `[impl Into<String>]`, a non-copy slice
   --> src/value.rs:139:21
    |
139 |         let value = label_values[i].into();
    |                     ^^^^^^^^^^^^^^^
    |                     |
    |                     cannot move out of here
    |                     move occurs because `label_values[_]` has type `impl Into<String>`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0508`.
error: could not compile `prometheus`.

To learn more, run the command again with --verbose.

**Comment by mxinden** [comment]
Digging deeper into this, I should have put more thoughts into my initial comment. Sorry @AlexanderThaller.

The `Into:into` trait method takes ownership of `self`:

```rust
pub trait Into<T> {
    fn into(self) -> T;
}
```

This is fine when passing single value to a function, e.g.:

```rust
foo(s: impl Into<String>)`
```

Problem is when one wants to pass not a single value, but multiple ones. So far we have done so by passing a slice. Problem is that the values passed via the slice by definition are passed as borrowed and not as owned values. Thus one can not call `Into::into` on the values.

Instead one could use the `ToString` trait which takes a reference `&self` instead of an owned `self`.

```rust
pub trait ToString {
    fn to_string(&self) -> String;
}
```

When the user passes an `&str` this is perfect as (a) its clear that the function is allocating and (b) its ergonomic as a user can just pass `&str` directly. Problem is once a user passes a `String`, while it is ergonomic, the implementation of `ToString` of `String` clones the underlying memory even though it is not needed in this case the user already passes an owned value (`String`).

As another alternative one could require a user to pass something implementing `ExactSizeIterator<Item = String>`. I don't think that would be ergonomic as all users previously doing `some_metric.with_label_values(&["value"]).inc()` would then need to do something along the lines of `with_label_values(["test".to_string()].into_iter().cloned())`.

We could explore using `Cow`, allowing one to pass both owned as well as borrowed strings, but that missed the point like `Into<String>` as when passed via a slice one does not have ownership of the `Cow` and thus would have to clone even when the value is `Cow::Owned`.

All in all I am unable to come up with a better solution (it is also quite late here in Germany) than the initial proposal of taking an `impl AsRef<str>`. Ideas welcome. Even though `&[impl AsRef<str>]` might not be particularly efficient in the case of a user having an owned `String` to spare, it is for sure ergonomic. I guess performance sensitive users should not use the crate itself in the first place, but the [`static-metrics` wrapper](https://github.com/tikv/rust-prometheus/tree/master/static-metric).

@AlexanderThaller what do you think? Again sorry for leading into the wrong direction.

AlexanderThaller:
No worries I'm always glad if people have ideas on improving my code (we also live in the same timezone also from germany :D). I think the idea of ergonomics vs efficiency is valid. We can maybe write documentation instead of using a type that communicates the allocating behavior but I'm unsure if that's the better option.

I will think about what we can do for now but maybe somebody has a better idea until then.

PR #310, opened 2020-03-17
use coarsetime::Instant::recent() instead

Renkai:
Use `recent()` instead of `now()` to improve performance, effect will be seen in https://github.com/tikv/tikv/pull/6698


sticnarf:
@breeswish From Rust 1.42, 32-bit macOS targets are not supported. We may not test them on CI.

PR #314, opened 2020-04-05
src/histogram: Make Histogram::observe atomic across collects

mxinden:
### Motivation

A histogram supports two main execution paths:

1. `observe` which increases the overall observation counter, updates the observation sum and increases a single bucket counter.

2. `collect` which snapshots the state of the histogram and exposes it as a Protobuf struct.

If an observe and a collect operation interleave, the latter could be exposing a snapshot of the histogram that does not uphold all histogram invariants. For example for the invariant that the overall observation counter should equal the sum of all bucket counters: Say that an `observe` increases the overall counter but before updating a specific bucket counter a `collect` operation snapshots the histogram.

The above race condition has been solved in the Golang Prometheus client with https://github.com/prometheus/client_golang/pull/457 by introducing the notion of shards, one hot shard for `observe` operations to record their observation and one cold shard for collect operations to collect a consistent snapshot of the histogram.

`observe` operations hit the hot shard and record their observation. Collect operations switch hot and cold, wait for all `observe` calls to finish on the previously hot now cold shard and then expose the consistent snapshot.

This pull request ports https://github.com/prometheus/client_golang/pull/457 to the Rust Prometheus client.

### Content of the pull request

The pull request contains three commits:

- `src/histogram: Add test ensuring Histogram::observe is atomic`

    Showcasing the above race condition in the current imlementation.

- `src/{histogram,atomic64}: Make Histogram::observe atomic across collects`

    Porting https://github.com/prometheus/client_golang/pull/457 to fix the race
    condition.

- `benches/histogram: Add benchmark for concurrent observe and collect`

    Adding a benchmark to show the impact of the patch. While the benchmark does not show a performance impact through the patch on my laptop, I am happy to test this more thoroughly on  a larger machine (128 cores) in case there is general interest to accept this patch.

### Greater picture

Fixing this race condition is especially attractive now that with Prometheus `v2.17.0` the isolation level has been increased (See changelog entry below and https://github.com/prometheus/prometheus/pull/6841).

> This release implements isolation in TSDB. API queries and recording rules are
> guaranteed to only see full scrapes and full recording rules.

### Trade-off

While this pull request fixes the above described race condition it does increase complexity:

- Introduction of the notion of shards.

- The `observe` code path is mostly untouched other than one additional atomic operation and increased `Ordering` levels and thus stays lock-free.

- `collect` operations need to happen sequentially (enforced through a single Mutex). A single `collect` and multiple `observe` operations can still operate concurrently. Given that the `collect` operation should happen rarely (> 1s) this should not introduce a performance impact.

- In order to coordinate hot and cold shards the 64 bit histogram counter is split into a 1 bit shard index and 63 bit counter. Thus the amount of observations a histogram can record is divided by two. While this might sound like an issue, one could still record one observation per milisecond for 292_277_266 years.

---

I hope the above description makes sense. Let me know if this is something you are willing to accept into `master`.

Thanks a bunch for maintaining this library!

### DIFF
--- a/src/atomic64.rs
+++ b/src/atomic64.rs
@@ -127,8 +127,8 @@ impl Atomic for AtomicF64 {
 
 impl AtomicF64 {
     /// Store the value, returning the previous value.
-    pub fn swap(&self, val: f64) -> f64 {
-        u64_to_f64(self.inner.swap(f64_to_u64(val), Ordering::Relaxed))
+    pub fn swap(&self, val: f64, ordering: Ordering) -> f64 {
+        u64_to_f64(self.inner.swap(f64_to_u64(val), ordering))
     }
 }
 
@@ -217,8 +217,8 @@ impl AtomicU64 {
 
     /// Swap the current value with the given value, returning that previously
     /// current value.
-    pub fn swap(&self, val: u64) -> u64 {
-        self.inner.swap(val, Ordering::Relaxed)
+    pub fn swap(&self, val: u64, ordering: Ordering) -> u64 {
+        self.inner.swap(val, ordering)
     }
 }
 
--- a/src/atomic64.rs
+++ b/src/atomic64.rs
@@ -125,6 +125,13 @@ impl Atomic for AtomicF64 {
     }
 }
 
+impl AtomicF64 {
+    /// Store the value, returning the previous value.
+    pub fn swap(&self, val: f64) -> f64 {
+        u64_to_f64(self.inner.swap(f64_to_u64(val), Ordering::Relaxed))
+    }
+}
+
 /// A atomic signed integer.
 #[derive(Debug)]
 pub struct AtomicI64 {
@@ -188,12 +195,30 @@ impl Atomic for AtomicU64 {
 
     #[inline]
     fn inc_by(&self, delta: Self::T) {
-        self.inner.fetch_add(delta, Ordering::Relaxed);
+        self.inc_by_with_ordering(delta, Ordering::Relaxed);
     }
 
     #[inline]
     fn dec_by(&self, delta: Self::T) {
         self.inner.fetch_sub(delta, Ordering::Relaxed);
+    }
+}
+
+impl AtomicU64 {
+    /// Get the value with the provided memory ordering.
+    pub fn compare_and_swap(&self, current: u64, new: u64, ordering: Ordering) -> u64 {
+        self.inner.compare_and_swap(current, new, ordering)
+    }
+
+    /// Increment the value by a given amount with the provided memory ordering.
+    pub fn inc_by_with_ordering(&self, delta: u64, ordering: Ordering) {
+        self.inner.fetch_add(delta, ordering);
+    }
+
+    /// Swap the current value with the given value, returning that previously
+    /// current value.
+    pub fn swap(&self, val: u64) -> u64 {
+        self.inner.swap(val, Ordering::Relaxed)
     }
 }
 
--- a/src/atomic64.rs
+++ b/src/atomic64.rs
@@ -215,8 +215,7 @@ impl AtomicU64 {
         self.inner.fetch_add(delta, ordering);
     }
 
-    /// Swap the current value with the given value, returning that previously
-    /// current value.
+    /// Stores a value into the atomic integer, returning the previous value.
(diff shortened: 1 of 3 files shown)


lucab:
@BusyJay @breeswish can you have a look at this PR?

I was chatting with @mxinden trying to figure out a better approach instead of `Mutex<()>`, but we ran out of smarter alternatives. Do you have any feedback on this? Or perhaps is this fine?

breeswish:
Sorry I missed this PR .. 🧐 @BusyJay Do you have suggestions about the Mutex in this PR?

BusyJay:
No, the mutex seems simple and reasonable.

src/histogram.rs
```diff
@@ -450,6 +450,8 @@ impl HistogramCore {
         hot_shard.count.inc_by(overall_count);
         hot_shard.sum.inc_by(cold_shard_sum);
 
+        drop(_collect_guard);
```
lucab, line 453: Given the variable is now used here, I don't think it needs a `_` prefix anymore, right?

mxinden: :+1: Not sure what the convention is. https://github.com/tikv/rust-prometheus/pull/314/commits/b0ef4e2682f0bd31f178bdb4715517232ab8dd72 removes the underscore.

mxinden:
@lucab @breeswish @BusyJay any objections in regards to this pull request? If not, would one of you mind approving it?

src/atomic64.rs
```diff
@@ -197,6 +204,24 @@ impl Atomic for AtomicU64 {
...
+    }
+
+    /// Increment the value by a given amount with the provided memory ordering.
+    pub fn inc_by_with_ordering(&self, delta: u64, ordering: Ordering) {
+        self.inner.fetch_add(delta, ordering);
+    }
+
+    /// Swap the current value with the given value, returning that previously
+    /// current value.
+    pub fn swap(&self, val: u64) -> u64 {
```
lucab, line 220: Should this take the ordering as an argument too?

src/atomic64.rs
```diff
@@ -197,6 +204,24 @@ impl Atomic for AtomicU64 {
...
+    pub fn compare_and_swap(&self, current: u64, new: u64, ordering: Ordering) -> u64 {
+        self.inner.compare_and_swap(current, new, ordering)
+    }
+
+    /// Increment the value by a given amount with the provided memory ordering.
+    pub fn inc_by_with_ordering(&self, delta: u64, ordering: Ordering) {
+        self.inner.fetch_add(delta, ordering);
+    }
+
+    /// Swap the current value with the given value, returning that previously
```
lucab, line 218: s/returning that/returning the/

src/histogram.rs
```diff
@@ -151,16 +154,176 @@ impl From<Opts> for HistogramOpts {
...
+            ShardIndex::First => 0,
+            ShardIndex::Second => 1,
+        }
+    }
+}
+
+/// An atomic u64 with the most significant used as a [`ShardIndex`] and the
+/// remaining 63 bits used to count [`Histogram`] observations.
+#[derive(Debug)]
+pub struct ShardAndCount {
```
lucab, line 228: Does this need to be `pub` or is perhaps `pub(crate)` enough?

src/histogram.rs
```diff
@@ -178,61 +341,131 @@ impl HistogramCore {
...
+
+            shard_and_count: ShardAndCount::new(),
+            shards: [Shard::new(buckets.len()), Shard::new(buckets.len())],
+
             upper_bounds: buckets,
-            counts,
         })
     }
 
+    /// Observe a given observation (f64) in the histogram.
```
lucab, line 357: s/Observe/Record/ perhaps?

src/atomic64.rs
mxinden: :+1: copied the `std` doc comment now.

https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html#method.swap

src/histogram.rs
mxinden: No need for being public at all. Good catch, thanks!

mxinden:
Thanks for the review @lucab. Would you mind taking another look?

PR #323, opened 2020-05-16
Added hygine to macros

Kixiron:
Fixes the macro hygiene problems with the macros when not using `use prometheus::*;`


breeswish:
@Kixiron Thanks for the contribution! You will need to sign-off **all of your commits** to pass the DCO check.

@nrc This PR contains a change that is generated by later rust-protobuf. Could you help check out whether it is compatible with TiKV?

breeswish:
@Kixiron Thanks for the contribution! You will need to sign-off **all of your commits** to pass the DCO check.

@nrc This PR contains a change that is generated by later rust-protobuf. Could you help check out whether it is compatible with TiKV?

PR #326, opened 2020-05-25
src/value: Have make_label_pairs error instead of panic

mxinden:
`make_label_pairs` expects `desc.variable_labels.len()` to equal the provided
`label_values.len()`. Instead of panicing on out of range indexing
return an error.

Fixes https://github.com/tikv/rust-prometheus/issues/306.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -44,14 +44,7 @@ impl<P: Atomic> Value<P> {
         label_values: &[&str],
     ) -> Result<Self> {
         let desc = describer.describe()?;
-        if desc.variable_labels.len() != label_values.len() {
-            return Err(Error::InconsistentCardinality {
-                expect: desc.variable_labels.len(),
-                got: label_values.len(),
-            });
-        }
-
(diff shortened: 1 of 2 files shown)


src/value.rs
```diff
@@ -44,14 +44,7 @@ impl<P: Atomic> Value<P> {
         label_values: &[&str],
     ) -> Result<Self> {
         let desc = describer.describe()?;
-        if desc.variable_labels.len() != label_values.len() {
```
mxinden, line 47: Moved into `make_label_pairs`.

breeswish:
Thanks!

CI is failed:

```
---- histogram::tests::test_error_on_inconsistent_label_cardinality stdout ----
thread 'histogram::tests::test_error_on_inconsistent_label_cardinality' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `1`', src/histogram.rs:1242:13
failures:
    histogram::tests::test_error_on_inconsistent_label_cardinality
```

Also you need to rebase all of your commits to sign-off in order to pass the DCO check.

PR #327, opened 2020-05-27
src/encoder/text: Optimize metric formatting

mxinden:
This pull request addresses the low-hanging-fruits within `src/encoder/text.rs` with the following commits:

- src/benches&src/bin: Add basic text encoder benchmark

- src/encoder/text: Use `write_all` instead of `write_fmt`

  Instead of going through `write!` which calls `write_fmt` use `write_all` directly skipping the additional formatting overhead.

- src/encoder/text: Use regex crate to determine if escaping is needed

  Within `escape_string` instead of always allocating a new String and iterating through each character one-by-one, first check whether the input string needs any escaping, if not, return without any allocation. If it does need escaping, start from the first occurence.

Benchmark results on a `Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz`:

```bash
$ cargo benchcmp master optimized
 name                                 master ns/iter  optimized ns/iter  diff ns/iter   diff %  speedup 
 bench_text_encoder_with_escaping     131,884,199     105,675,740         -26,208,459  -19.87%   x 1.25 
 bench_text_encoder_without_escaping  117,592,719     73,070,167          -44,522,552  -37.86%   x 1.61
```

CPU flamegraph before:
master

CPU flamegraph after:
optimized

Memory allocation flamegraph before:
image

Memory allocation flamegraph after:
image

### DIFF
--- a/src/encoder/text.rs
+++ b/src/encoder/text.rs
@@ -32,23 +32,33 @@ impl Encoder for TextEncoder {
             // Fail-fast checks.
             check_metric_family(mf)?;
 
+            // Write `# HELP` header.
             let name = mf.get_name();
             let help = mf.get_help();
             if !help.is_empty() {
-                writeln!(writer, "# HELP {} {}", name, escape_string(help, false))?;
+                writer.write_all(b"# HELP ")?;
+                writer.write_all(name.as_bytes())?;
+                writer.write_all(b" ")?;
+                writer.write_all(escape_string(help, false).as_bytes())?;
+                writer.write_all(b"\n")?;
             }
 
+            // Write `# TYPE` header.
             let metric_type = mf.get_field_type();
             let lowercase_type = format!("{:?}", metric_type).to_lowercase();
-            writeln!(writer, "# TYPE {} {}", name, lowercase_type)?;
+            writer.write_all(b"# TYPE ")?;
+            writer.write_all(name.as_bytes())?;
+            writer.write_all(b" ")?;
+            writer.write_all(lowercase_type.as_bytes())?;
+            writer.write_all(b"\n")?;
 
             for m in mf.get_metric() {
                 match metric_type {
                     MetricType::COUNTER => {
-                        write_sample(name, m, "", "", m.get_counter().get_value(), writer)?;
+                        write_sample(writer, name, None, m, None, m.get_counter().get_value())?;
                     }
                     MetricType::GAUGE => {
-                        write_sample(name, m, "", "", m.get_gauge().get_value(), writer)?;
+                        write_sample(writer, name, None, m, None, m.get_gauge().get_value())?;
                     }
                     MetricType::HISTOGRAM => {
                         let h = m.get_histogram();
@@ -57,12 +67,12 @@ impl Encoder for TextEncoder {
                         for b in h.get_bucket() {
                             let upper_bound = b.get_upper_bound();
                             write_sample(
-                                &format!("{}_bucket", name),
-                                m,
-                                BUCKET_LABEL,
-                                &format!("{}", upper_bound),
-                                b.get_cumulative_count() as f64,
                                 writer,
+                                name,
+                                Some("_bucket"),
+                                m,
+                                Some((BUCKET_LABEL, &upper_bound.to_string())),
+                                b.get_cumulative_count() as f64,
                             )?;
                             if upper_bound.is_sign_positive() && upper_bound.is_infinite() {
                                 inf_seen = true;
@@ -70,31 +80,24 @@ impl Encoder for TextEncoder {
                         }
                         if !inf_seen {
                             write_sample(
-                                &format!("{}_bucket", name),
-                                m,
-                                BUCKET_LABEL,
-                                POSITIVE_INF,
-                                h.get_sample_count() as f64,
                                 writer,
+                                name,
+                                Some("_bucket"),
+                                m,
+                                Some((BUCKET_LABEL, POSITIVE_INF)),
+                                h.get_sample_count() as f64,
                             )?;
                         }
 
-                        write_sample(
-                            &format!("{}_sum", name),
-                            m,
-                            "",
-                            "",
-                            h.get_sample_sum(),
-                            writer,
-                        )?;
+                        write_sample(writer, name, Some("_sum"), m, None, h.get_sample_sum())?;
 
                         write_sample(
-                            &format!("{}_count", name),
-                            m,
-                            "",
-                            "",
-                            h.get_sample_count() as f64,
                             writer,
+                            name,
+                            Some("_count"),
+                            m,
+                            None,
+                            h.get_sample_count() as f64,
                         )?;
                     }
                     MetricType::SUMMARY => {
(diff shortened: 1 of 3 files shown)


src/encoder/text.rs
```diff
@@ -32,23 +34,33 @@ impl Encoder for TextEncoder {
...
+            // Write `# HELP` header.
             let name = mf.get_name();
             let help = mf.get_help();
             if !help.is_empty() {
-                writeln!(writer, "# HELP {} {}", name, escape_string(help, false))?;
+                writer.write_all(b"# HELP ")?;
+                writer.write_all(name.as_bytes())?;
+                writer.write_all(b" ")?;
+                writer.write_all(escape_string(help, false).as_bytes())?;
+                writer.write_all(b"\n")?;
```
lucab, line 45: This pattern looks like a [vectored write](https://en.wikipedia.org/wiki/Vectored_I/O). Did you consider directly going for https://doc.rust-lang.org/std/io/trait.Write.html#method.write_vectored here?

src/encoder/text.rs
```diff
@@ -224,26 +218,49 @@ fn label_pairs_to_text(
 
 /// `escape_string` replaces `\` by `\\`, new line character by `\n`, and `"` by `\"` if
 /// `include_double_quote` is true.
-fn escape_string(v: &str, include_double_quote: bool) -> String {
-    let mut escaped = String::with_capacity(v.len() * 2);
+///
+/// Implementation adapted from
+/// https://lise-henry.github.io/articles/optimising_strings.html
```
lucab, line 223: I somehow have the hope that regex `replace_all` performance improved over that 2016 page, as it changed in the meanwhile to work on Cow. Did you try benchmarking both here?

src/encoder/text.rs
kennytm: `write_vectored` won't guarantee that everything in the `&[IoSlice]` are written out, unlike `write_all`. In fact, the default implementation of `write_vectored` on non-files will only write out the first non-empty slice. This means these 5 lines will need to become a loop if we use `write_vectored` instead.

src/encoder/text.rs
mxinden: I hoped as well, but the diff below degrades performance extensively:

```bash
$ cargo benchcmp optimized optimized_replace_all 
 name                                 optimized ns/iter  optimized_replace_all ns/iter  diff ns/iter   diff %  speedup 
 bench_text_encoder_with_escaping     105,675,740        253,291,262                     147,615,522  139.69%   x 0.42 
 bench_text_encoder_without_escaping  73,070,167         90,540,884                       17,470,717   23.91%   x 0.81
```

Am I missing something?

```diff
diff --git a/src/encoder/text.rs b/src/encoder/text.rs
index 97a5ca5..f8179aa 100644
--- a/src/encoder/text.rs
+++ b/src/encoder/text.rs
@@ -1,6 +1,6 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
-use regex::{Match, Regex};
+use regex::{Captures, Match, Regex};
 use std::borrow::Cow;
 use std::io::Write;
 
@@ -218,9 +218,6 @@ fn label_pairs_to_text(
 
 /// `escape_string` replaces `\` by `\\`, new line character by `\n`, and `"` by `\"` if
 /// `include_double_quote` is true.
-///
-/// Implementation adapted from
-/// https://lise-henry.github.io/articles/optimising_strings.html
 fn escape_string(v: &str, include_double_quote: bool) -> Cow<'_, str> {
     // Regex compilation is expensive. Use `lazy_static` to compile the regexes
     // once per process lifetime and not once per function invocation.
@@ -229,37 +226,18 @@ fn escape_string(v: &str, include_double_quote: bool) -> Cow<'_, str> {
         static ref QUOTED_ESCAPER: Regex = Regex::new("(\\\\|\n|\")").expect("Regex to be valid.");
     }
 
-    let first_occurence = if include_double_quote {
-        QUOTED_ESCAPER.find(v)
-    } else {
-        ESCAPER.find(v)
-    }
-    .as_ref()
-    .map(Match::start);
-
-    if let Some(first) = first_occurence {
-        let mut escaped = String::with_capacity(v.len() * 2);
-        escaped.push_str(&v[0..first]);
-        let remainder = v[first..].chars();
-
-        for c in remainder {
-            match c {
-                '\\' | '\n' => {
-                    escaped.extend(c.escape_default());
-                }
-                '"' if include_double_quote => {
-                    escaped.extend(c.escape_default());
-                }
-                _ => {
-                    escaped.push(c);
-                }
-            }
+    let replacer = |cap: &Captures| {
+        match &cap[0] {
+            c@ "\\" | c@"\n" => c.escape_default().to_string(),
+            c@"\"" if include_double_quote => c.escape_default().to_string(),
+            c @ _ => panic!("Got unexpected capture {:?}.", c),
         }
-        escaped.into()
+    };
+
+    if include_double_quote {
+        QUOTED_ESCAPER.replace_all(v, replacer)
     } else {
-        // The input string does not contain any characters that would need to
-        // be escaped. Return it as it is.
-        v.into()
+        ESCAPER.replace_all(v, replacer)
     }
 }
```

src/encoder/text.rs
mxinden: For `<Vec as Write>` it would have the benefit of reserving enough space for all slices at once: 

```rust
    #[inline]
    fn write_vectored(&mut self, bufs: &[IoSlice<'_>]) -> io::Result<usize> {
        let len = bufs.iter().map(|b| b.len()).sum();
        self.reserve(len);
        for buf in bufs {
            self.extend_from_slice(buf);
        }
        Ok(len)
    }
```

https://doc.rust-lang.org/src/std/io/impls.rs.html#337-364

Not sure how to handle the case where not everything was written at once efficiently yet. Will need to put more thoughts into that. Not sure whether we just want to copy https://github.com/rust-lang/rust/pull/70612/.

lucab: @kennytm I missed that nuance, thanks for sharing that!

src/encoder/text.rs
lucab: I tried some further changes to the replacer, but I never got any better than 0.58x than your PR.

src/encoder/text.rs
```diff
@@ -224,26 +218,49 @@ fn label_pairs_to_text(
...
             }
         }
+        escaped.into()
+    } else {
+        // The input string does not contain any characters that would need to
+        // be escaped. Return it as it is.
+        v.into()
     }
-
-    escaped.shrink_to_fit();
```
lucab, line 244: This is still relevant for the `Some(first)` arm, right?

src/encoder/text.rs
lucab: @mxinden I'm fine deferring this pending libstd stabilization.

src/encoder/text.rs
mxinden: Out of curiosity, what were your changes @lucab?

lucab: @mxinden dropping the group/capturing in regex and plain `str` replacements arms (like `"\\" =>  "\\\\"`). Sorry I've already thrown away that branch.

src/encoder/text.rs
mxinden: Ah, good catch. Sorry for missing that. Speeds up the escaping test a bit:

```bash
$ cargo benchcmp optimized optimized_shrink                                   
 name                                 optimized ns/iter  optimized_shrink ns/iter  diff ns/iter  diff %  speedup    
 bench_text_encoder_with_escaping     105,675,740        99,516,087                  -6,159,653  -5.83%   x 1.06    
 bench_text_encoder_without_escaping  73,070,167         72,932,821                    -137,346  -0.19%   x 1.00
```
Fix included in https://github.com/tikv/rust-prometheus/pull/327/commits/93b35e23a658de3888501c6da3b1712e5137b94e.

mxinden:
@lucab @kennytm would you mind giving this another review?

lucab:
LGTM, but leaving some room for others' feedback.

breeswish:
It's interesting that `write!` is not efficient as I thought. @kennytm @nrc PTAL

kennytm:
@breeswish yes `std::fmt` is not efficient [because of the virtual calls](https://internals.rust-lang.org/t/changing-core-fmt-for-speed/5154), which is why packages like [ufmt](https://crates.io/crates/ufmt) exists.

mxinden:
Any further comments on this patch?

@breeswish in case there are no objections, do you want me to merge my own pull-request, or rather always have someone other than the author click the merge button?

mxinden:
I am guessing there are no objections to this patch. @lucab @breeswish could one of you approve this pull request?

lucab:
@mxinden stamped, I think you can rebase-squash a couple of commits from this branch though. Feel free to self-merge when done.

breeswish:
@lucab Feel free to just merge PRs :)

PR #328, opened 2020-06-11
proto/proto_model: Run rust fmt

mxinden:
Currently all CI builds are failing due to Rust fmt trying to format `proto/proto_model.rs`.

For now this patch simply formats the given file.

In the future I would suggest:

1. Have rustfmt ignore `proto/` via the (not yet stable) `ignore` attribute in `rustfmt.toml`.

2. Generate `proto/proto_model.rs` on each CI run to ensure the checked-in version matches and is up-to-date.

### DIFF
--- a/proto/proto_model.rs
+++ b/proto/proto_model.rs
@@ -21,7 +21,7 @@
 use protobuf::Message as Message_imported_for_functions;
 use protobuf::ProtobufEnum as ProtobufEnum_imported_for_functions;
 
-#[derive(PartialEq,Clone,Default)]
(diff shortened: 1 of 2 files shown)


lucab:
@mxinden what about throwing a `#[rustfmt::skip]` in front of the mod declaration instead? That one should work on stable already.

mxinden:
@lucab for sure the better solution. Adjusted accordingly. Thanks.

lucab:
@mxinden feel free to squash and self-merge.

PR #334, opened 2020-07-05
process: Add swap memory metrics

quodlibetor:
Related to rss and virtual size, this adds swapped-out memory.

### DIFF
--- a/src/process_collector.rs
+++ b/src/process_collector.rs
@@ -16,7 +16,7 @@ use crate::proto;
 pub use libc::pid_t;
 
 // Six metrics per ProcessCollector.
-const MERTICS_NUMBER: usize = 6;
+const MERTICS_NUMBER: usize = 7;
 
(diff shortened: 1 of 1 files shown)


breeswish:
Thanks for the contribution! The process collector is intentionally to be identical to the official [Golang client implementation](https://github.com/prometheus/client_golang/blob/040c98c8e638eb358dfc1b4b04b385a8df21e6a7/prometheus/process_collector.go). If you would like to report swap memory, maybe it is better to introduce it at the application side?

lucab:
However the `METRICS` typo fix would be to nice to land. @quodlibetor can you please split that to its own PR?

quodlibetor:
Ah, I didn't realize that there was some other client that this was trying to match. Makes sense!

quodlibetor:
> However the METRICS typo fix would be to nice to land.

Moved it to #335

PR #335, opened 2020-07-06
Fix typo in constant

quodlibetor:


src/process_collector.rs
```diff
@@ -165,7 +165,7 @@ impl Collector for ProcessCollector {
         };
 
         // collect MetricFamilys.
-        let mut mfs = Vec::with_capacity(MERTICS_NUMBER);
+        let mut mfs = Vec::with_capacity(METRICS_NUMBER);
         mfs.extend(cpu_total_mfs);
         mfs.extend(self.open_fds.collect());
         mfs.extend(self.max_fds.collect());
```
mxinden, line 171: [As far as I can tell `vec![]` only allocates once.](https://doc.rust-lang.org/src/alloc/macros.rs.html#46) See also [`into_vec`](https://doc.rust-lang.org/std/primitive.slice.html#method.into_vec).

As the length is statically defined at compile time, the following would work as well, no? Would relief us from the possibility of the constant ever being outdated.

```suggestion
        let mut mfs = vec![
          mfs.extend(cpu_total_mfs);
          mfs.extend(self.open_fds.collect());
          mfs.extend(self.max_fds.collect());
```

quodlibetor: your suggestion mutates `mfs` inside the macro that is creating it, if I'm reading it correctly, which doesn't make sense: `let mfs = vec![mfs.extend(..), mfs.extend(..)];`?

The constant being correct is currently verified inside a test, and as mentioned in the linked PR it is unlikely to change much since this library is trying to exactly match the facts from the go version of `process_collector`.

mxinden: @quodlibetor absolutely right. Should have put more thoughts into my suggestion. Sorry for the noise.

PR #343, opened 2020-09-03
Fix #342: Async support for push_metrics

### LINKED ISSUE #342: Push doesn't work inside tokio async block
**Describe the bug**
The `prometheus::push_metrics` function panics if called inside an async block (at least when using the Tokio runtime). It panics with the message 'Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.'

I believe this is due to the use of `reqwest::blocking::Client` in the implementation. One backwards-compatible solution would be adding a new `prometheus::push_metrics_async` function that uses `reqwest::Client` instead. I would be happy to open a PR to add this, if you like the suggestion.

**To Reproduce**
Minimal example to reproduce the behavior:

src/main.rs
```rust
#[macro_use]
extern crate prometheus;
use prometheus::{gather, push_metrics, Error};

#[tokio::main]
async fn main() -> Result<(), Error> {
    let counter = register_counter!("example_counter", "Just a counter").unwrap();
    counter.inc();
    push_metrics("example", HashMap::new(), "localhost:9091", gather(), None)?;
    println!("Success");
    Ok(())
}
```

Cargo.toml
```toml
[package]
name = "prometheus-bug"
version = "0.1.0"
authors = ["Adam Chalmers <adam.s.chalmers@gmail.com>"]
edition = "2018"

[dependencies]
tokio = { version = "0.2.22", features = ["macros"] }

[dependencies.prometheus]
version = "0.10"
default-features = true
features = ["push", "process"]
```

**Expected behavior**
`push_metrics()` does not panic when called inside an async function using Tokio.

**System information**
* CPU architecture: Intel i7
* Distribution and kernel version: MacOS Mojave

adamchalmers:
I tried to document the problem thoroughly on the issue itself. This is a proposed solution. I tried to share as much code as I could between `push_async` and `push`. Let me know if you'd rather I take a different approach!

Fixes: #342

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -223,7 +223,8 @@ pub use self::histogram::{Histogram, HistogramOpts, HistogramTimer, HistogramVec
 pub use self::metrics::Opts;
 #[cfg(feature = "push")]
 pub use self::push::{
-    hostname_grouping_key, push_add_collector, push_add_metrics, push_collector, push_metrics,
+    hostname_grouping_key, push_add_collector, push_add_collector_async, push_add_metrics,
+    push_add_metrics_async, push_collector, push_collector_async, push_metrics, push_metrics_async,
     BasicAuthentication,
 };
 pub use self::registry::Registry;
--- a/src/push.rs
+++ b/src/push.rs
@@ -18,13 +18,6 @@ use crate::registry::Registry;
 
 const REQWEST_TIMEOUT_SEC: Duration = Duration::from_secs(10);
 
-lazy_static! {
-    static ref HTTP_CLIENT: Client = Client::builder()
-        .timeout(REQWEST_TIMEOUT_SEC)
-        .build()
-        .unwrap();
-}
-
 /// `BasicAuthentication` holder for supporting `push` to Pushgateway endpoints
 /// using Basic access authentication.
 /// Can be passed to any `push_metrics` method.
@@ -60,6 +53,18 @@ pub fn push_metrics<S: BuildHasher>(
     push(job, grouping, url, mfs, "PUT", basic_auth)
 }
 
+/// Functions just like `push_metrics`, except the metrics are pushed
+/// asynchronously.
+pub async fn push_metrics_async<S: BuildHasher>(
+    job: &str,
+    grouping: HashMap<String, String, S>,
+    url: &str,
+    mfs: Vec<proto::MetricFamily>,
+    basic_auth: Option<BasicAuthentication>,
+) -> Result<()> {
+    push_async(job, grouping, url, mfs, "PUT", basic_auth).await
+}
+
 /// `push_add_metrics` works like `push_metrics`, but only previously pushed
 /// metrics with the same name (and the same job and other grouping labels) will
 /// be replaced. (It uses HTTP method 'POST' to push to the Pushgateway.)
@@ -73,16 +78,25 @@ pub fn push_add_metrics<S: BuildHasher>(
     push(job, grouping, url, mfs, "POST", basic_auth)
 }
 
+/// `push_add_metrics_async` works like `push_metrics`, but async.
+pub async fn push_add_metrics_async<'a, S: BuildHasher>(
+    job: &'a str,
+    grouping: HashMap<String, String, S>,
+    url: &'a str,
+    mfs: Vec<proto::MetricFamily>,
+    basic_auth: Option<BasicAuthentication>,
+) -> Result<()> {
+    push(job, grouping, url, mfs, "POST", basic_auth)
+}
+
 const LABEL_NAME_JOB: &str = "job";
 
-fn push<S: BuildHasher>(
+fn configure_push<S: BuildHasher>(
     job: &str,
     grouping: HashMap<String, String, S>,
     url: &str,
     mfs: Vec<proto::MetricFamily>,
-    method: &str,
-    basic_auth: Option<BasicAuthentication>,
-) -> Result<()> {
+) -> Result<(String, impl Encoder, Vec<u8>)> {
     // Suppress clippy warning needless_pass_by_value.
     let grouping = grouping;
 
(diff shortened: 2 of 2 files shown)


src/push.rs
```diff
@@ -23,6 +24,10 @@ lazy_static! {
         .timeout(REQWEST_TIMEOUT_SEC)
         .build()
         .unwrap();
+    static ref ASYNC_HTTP_CLIENT: AsyncClient = AsyncClient::builder()
```
mxinden, line 27: In my eyes defining an HTTP client as a global static is problematic, for all the reasons global variables are. While that is somewhat subjective, creating both an async client and a sync client, even though a user likely only needs one, sounds even worse.

What do people think of requiring users to create a client on demand and binding `push` as a method to such client?

src/push.rs
```diff
@@ -60,6 +65,18 @@ pub fn push_metrics<S: BuildHasher>(
     push(job, grouping, url, mfs, "PUT", basic_auth)
 }
 
+/// Functions just like `push_metrics`, except the metrics are pushed
+/// asynchronously.
+pub async fn push_metrics_async<S: BuildHasher>(
```
mxinden, line 70: For the sake of consistency we should also offer the remaining `pub` functions in an asynchronous fashion, e.g. `push_async_collector`.

mxinden (commented):
Thanks for both tracing this bug all the way down as well as proposing a bug fix.

I am likely the wrong maintainer to review this pull request as (a) I am not a user of pushing metrics to a Prometheus server and (b) I am of the opinion that pushing metrics is out of scope for this library.

With that in mind I would prefer another maintainer to get involved as well.

mxinden:
Fixes https://github.com/tikv/rust-prometheus/issues/342

_Properly linking back to the detailed bug report._

src/push.rs
adamchalmers: I agree with you that the globals aren't ideal, but I feel like lazy_static is a very ergonomic solution for consumers of the library. I like the idea of the user instantiating a client, but it's got a few little usability woes:

- Users now have to explicitly depend on reqwest and familiarize themselves with it works, how to create a client, etc.
- It's a breaking change (I know this is still a 0.x crate, so breaking changes are expected, but we should try to minimize them)

What if instead, we just made async support a feature that users had to opt into? That way we could instantiate the async client if and only if the feature is enabled?

src/push.rs
adamchalmers: Agreed, done!

adamchalmers:
Tagging @koushiro @breeswish for review :)

src/push.rs
lucab: I fear than adding an additional either-or feature is only making the whole crate harder to test and maintain.

README.md
```diff
@@ -25,7 +25,9 @@ This crate provides several optional components which can be enabled via [Cargo
 
 - `process`: Enable [process metrics](https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics) support.
 
-- `push`: Enable [push metrics](https://prometheus.io/docs/instrumenting/pushing/) support.
+- `push`: Enable [push metrics](https://prometheus.io/docs/instrumenting/pushing/) support with blocking network calls. Incompatible with `push-async`.
```
mxinden, line 28: > Incompatible with `push-async`.

Keep in mind that Rust features are additive, thus no feature should be incompatible with another one.

https://github.com/rust-lang/cargo/issues/4328

breeswish: Just an idea: For simplicity, how about leaving push-async alone and remove push, considering that from async to sync is easy (just execute an executor block_on). Discussion is welcome!

adamchalmers: I'd be happy to do that! It just depends on what the maintainers of this project want -- I'm not sure how they balance breaking changes vs. updating to the latest and greatest. All my Rust projects are using async, so, selfishly, I'd be fine with that.

adamchalmers: Actually I just changed my PR. Now the async functions are exported in the `push` feature. I've addressed @mxinden's concern about instantiating lazy_statics unnecessarily by moving their instantiation to their call site. Now you'll only instantiate a reqwest client if you actually use it to push metrics. I think is an improvement!

src/push.rs
adamchalmers: Agreed. I've removed the feature now.

adamchalmers:
@mxinden @breeswish Hi! Just pinging y'all for another review :)

README.md
```diff
@@ -25,7 +25,7 @@ This crate provides several optional components which can be enabled via [Cargo
 
 - `process`: Enable [process metrics](https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics) support.
 
-- `push`: Enable [push metrics](https://prometheus.io/docs/instrumenting/pushing/) support.
+- `push`: Enable [push metrics](https://prometheus.io/docs/instrumenting/pushing/) support with blocking network calls.
```
breeswish, line 28: The change here may be reverted

src/lib.rs
```diff
@@ -108,7 +108,7 @@ This library supports four features:
   using the pre-generated client.
 * `nightly`: Enable nightly only features.
 * `process`: For collecting process info.
-* `push`: Enable push support.
+* `push`: Enable push support using blocking network calls. Incompatible with `push`.
```
breeswish, line 111: We provide async push now, so maybe it can be reverted

src/push.rs
```diff
@@ -159,14 +190,50 @@ fn push<S: BuildHasher>(
     }
 
     let response = builder.send().map_err(|e| Error::Msg(format!("{}", e)))?;
+    handle_push_response(response.status(), push_url)
+}
+
+async fn push_async<S: BuildHasher>(
```
breeswish, line 196: This part is very similar to `fn push`. Is it possible to use something like https://docs.rs/crate/reqwest/0.10.8/source/src/blocking/wait.rs?

breeswish (commented):
Thanks for the contribution! I have left a few comments.

src/push.rs
adamchalmers: I don't understand what the linked code has to do with duplication. I tried to reduce duplication as much as possible by breaking common code into their own functions which are called inside both async_push and push. But this can only go so far, because the underlying reqwest structs use different types.

adamchalmers:
Sorry I forgot about this for so many months. I've addressed most of the feedback except the one comment I didn't understand.

adamchalmers:
Sorry I forgot about this for so many months. I've addressed most of the feedback except the one comment I didn't understand. I'd like to get this merged, because my project is stuck on Prometheus 0.7, which means I'm stuck pulling in an insecure version of Hyper (via reqwest), which means that my work's CI blocks builds. I can always fork this repo until the PR is merged, but I'd rather get this addressed :)

adamchalmers:
PR is no longer necessary because I realized that Tokio 1.0 doesn't clash with reqwest::blocking!

adamchalmers:
Actually, the latest reqwest 0.11 and the latest tokio 1 are incompatible again. I think this PR is ultimately the best way forward, that way projects which use rust-prometheus can make their own choice of async or blocking.

WaffleLapkin:
What's the status of this PR? It seems odd, to not provide an async interface when using `reqwest`. Calling blocking code (especially something like an HTTP client) from async is never a good idea.

ashlineldridge:
Any update on this PR? Would be great to this functionality land 🙏🏼

adamchalmers:
I agree, it would be great for one of the maintainers to review this and tell me what is blocking merge.

adamchalmers:
Closing as apparently no maintainers want this reviewed. Ping me if you want it reopened.

PR #349, opened 2020-09-17
Support adding prefix to metric name by env variable

solotzg:
For some reason, we need to batch add prefix to metric name rather than modifying code.
Implement it as a new feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ gen = ["protobuf-codegen-pure"]
 nightly = ["libc"]
 process = ["libc", "procfs"]
 push = ["reqwest", "libc", "protobuf"]
+metric-name-prefix=[]
 
 [dependencies]
 cfg-if = "^0.1"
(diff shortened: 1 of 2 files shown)


mxinden (changes requested):
> For some reason, we need to batch add prefix to metric name rather than modifying code.

@solotzg can you add more details what your use case is, especially why you can not modify the user code?

To me a library should be as [pure (in the functional sense)](https://en.wikipedia.org/wiki/Pure_function) as possible. Instead of introducing a possible side-effect through environment variables within the library itself, I think the name prefix, wherever it might come from, should be passed to the library as an argument.

lucab:
Same feedback as @mxinden from my side.

If for some reason the name itself can not be tweaked, it it possible to achieve the same effect by adding a [namespace](https://docs.rs/prometheus/0.10.0/prometheus/struct.Opts.html#method.namespace) to the metric or by setting a [prefix](https://docs.rs/prometheus/0.10.0/prometheus/struct.Registry.html#method.new_custom) on the registry.

PR #351, opened 2020-10-10
static-metric/src/parser: Allow not specifying visibility

mxinden:
When determining whether to parse a struct or an enum the `Parse`
implementation for `StaticMetricMacroBodyItem` checks for the tokens
`struct` or `enum`. Given that the two can optionally be preceeded by
a visibility token (e.g. `pub(crate)`) one needs to peek at both the
first and the second token.

Follow up on https://github.com/tikv/rust-prometheus/pull/268/.

### DIFF
--- a/static-metric/CHANGELOG.md
+++ b/static-metric/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog for prometheus-static-metric
 
+## 0.5.0 [unreleased]
+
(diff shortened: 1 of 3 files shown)


lucab:
@mxinden clippy has a few suggestions on this one.

mxinden:
Thanks for the pointer @lucab. Fixed in https://github.com/tikv/rust-prometheus/pull/354.

PR #353, opened 2020-10-10
.travis.yml: Run benchmarks on each run

mxinden:
Run benchmarks on each pull request to ensure they compile and won't panic. 

In the long run we might want to remove the line below fixing the nightly version instead.

```yaml
  allow_failures:
  - rust: nightly
```

Signed-off-by: Max Inden <mail@max-inden.de>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,4 +44,8 @@ script:
     if [ $TRAVIS_RUST_VERSION = 'nightly' ]; then
         cargo build -p prometheus-static-metric --examples --no-default-features --features="$FEATURES"
         cargo test -p prometheus-static-metric --no-default-features --features="$FEATURES"
+
+        # Run benchmarks.
+        cargo bench --no-default-features --features="$FEATURES"
+        cargo bench -p prometheus-static-metric --no-default-features --features="$FEATURES"
     fi


breeswish:
Oh, it already failed.. 😱

mxinden:
@breeswish addressed in https://github.com/tikv/rust-prometheus/pull/354.

breeswish (approved):

lucab:
Just wondering, in order to remove the nightly dependency for benchmarks, has anybody already tried [criterion](https://github.com/bheisler/criterion.rs) maybe? Any feedback/experience to share?

mxinden:
@lucab I have been using criterion in all my personal projects. Only drawback I see would be the small learning curve required at the beginning.

lucab:
@mxinden ack. I don't personally have much experience with benchmarks in general, so I can't add real feedback here. All pros&cons weighted, do you think it would make sense exploring at switch here? Perhaps the PingCAP folks have some input on this?

breeswish:
> @mxinden ack. I don't personally have much experience with benchmarks in general, so I can't add real feedback here. All pros&cons weighted, do you think it would make sense exploring a switch here? Perhaps the PingCAP folks have some input on this?

In PingCAP we used criterion heavily, [examples](https://github.com/tikv/tikv/blob/master/tests/benches/coprocessor_executors/mod.rs). Criterion is very suitable for writing serious benchmarks, for example, comparing different parameters, isolating initialization costs, etc. However PingCAP projects also heavily rely on nightly :) ..

PR #354, opened 2020-10-12
src/desc: Validate metric and label names via regex crate

mxinden:
Current CI runs fail (see https://github.com/tikv/rust-prometheus/pull/353) due to the following Clippy warning:

```
error: match expression looks like `matches!` macro
  --> src/desc.rs:19:16
   |
19 |               && match c as u8 {
   |  ________________^
20 | |                 b'a'..=b'z' | b'A'..=b'Z' | b'_' | b':' => true,
21 | |                 _ => false,
22 | |             }
   | |_____________^ help: try this: `matches!(c as u8, b'a'..=b'z' | b'A'..=b'Z' | b'_' | b':')`
   |
   = note: `-D clippy::match-like-matches-macro` implied by `-D clippy`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#match_like_matches_macro
```

Instead of using `matches!` I propose to remove the entire manual validation in favor of a simple regex.

For those worried about performance, even though I doubt this code path is performance crticial, I added a basic benchmark showing not a regression but a slight speed-up.

```bash
$ cargo benchcmp manual regex
 name                    manual ns/iter  regex ns/iter  diff ns/iter  diff %  speedup 
 description_validation  684             665                     -19  -2.78%   x 1.03

$ cat /proc/cpuinfo | grep "model name" | head -1
model name      : Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
```

### DIFF
--- a/benches/desc.rs
+++ b/benches/desc.rs
@@ -1,0 +1,31 @@
+// Copyright 2020 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![feature(test)]
+
+extern crate test;
+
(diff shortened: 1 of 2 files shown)


lucab (approved):

breeswish (approved):

breeswish:
Thanks!

PR #359, opened 2020-10-30
Fix example in with_label_values() function

davidor:


mxinden (commented):
Good catch. Thank you for the pull request. 

What do you think of making this a [Documentation test](https://doc.rust-lang.org/rustdoc/documentation-tests.html)? Thus we would catch similar bugs at compile time in the future.

davidor:
@mxinden I changed it to be a doc test :+1:

lucab:
LGTM, thanks for the fix!

lucab (approved):

PR #365, opened 2020-11-06
Change Integer Counter type from AtomicI64 to AtomicU64

### LINKED ISSUE #364: Integer Counter type uses signed type (AtomicI64), when unsigned might be preferrable?
`src/counter.rs` defines:

```rust
pub type IntCounter = GenericCounter<AtomicI64>;
```
Looking at the Prometheus spec:

"A counter is a cumulative metric that represents a single monotonically increasing counter whose value can only increase or be reset to zero on restart. For example, you can use a counter to represent the number of requests served, tasks completed, or errors."

I was wondering why `AtomicU64` hadn't been used instead, since according to the spec, negative values are invalid.

With the code as it is, a counter can overflow and wrap to a negative number:

```
# HELP overflow test integer counter
# TYPE overflowable counter
overflowable -6446744073709552000
```

... I think that Prometheus will gracefully handle a reset to zero (e.g. a wrapped AtomicU64), it won't gracefully handle a wrapped signed integer I think (`rate()` won't expect this) I think?  https://www.robustperception.io/how-does-a-prometheus-counter-work

Folyd:
According to the definition of Prometheus's [Counter type](https://prometheus.io/docs/concepts/metric_types/#counter), the `AtomicU64` is preferred to `AtomicI64` for `IntCounter*` types:
> A counter is a cumulative metric that represents a single monotonically increasing counter whose value can only increase or be reset to zero on restart. 

Fix #364 .

### DIFF
--- a/src/counter.rs
+++ b/src/counter.rs
@@ -6,7 +6,7 @@ use std::collections::HashMap;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
-use crate::atomic64::{Atomic, AtomicF64, AtomicI64, Number};
+use crate::atomic64::{Atomic, AtomicF64, AtomicU64, Number};
 use crate::desc::Desc;
 use crate::errors::Result;
 use crate::metrics::{Collector, LocalMetric, Metric, Opts};
@@ -25,7 +25,7 @@ pub type Counter = GenericCounter<AtomicF64>;
 
 /// The integer version of [`Counter`]. Provides better performance if metric values
 /// are all integers.
-pub type IntCounter = GenericCounter<AtomicI64>;
+pub type IntCounter = GenericCounter<AtomicU64>;
 
 impl<P: Atomic> Clone for GenericCounter<P> {
     fn clone(&self) -> Self {
@@ -142,7 +142,7 @@ pub type CounterVec = GenericCounterVec<AtomicF64>;
 
 /// The integer version of [`CounterVec`]. Provides better performance if metric
 /// values are all integers.
-pub type IntCounterVec = GenericCounterVec<AtomicI64>;
+pub type IntCounterVec = GenericCounterVec<AtomicU64>;
 
 impl<P: Atomic> GenericCounterVec<P> {
     /// Create a new [`GenericCounterVec`] based on the provided
@@ -186,7 +186,7 @@ pub type LocalCounter = GenericLocalCounter<AtomicF64>;
 
 /// The integer version of [`LocalCounter`]. Provides better performance
 /// if metric values are all integers.
-pub type LocalIntCounter = GenericLocalCounter<AtomicI64>;
+pub type LocalIntCounter = GenericLocalCounter<AtomicU64>;
 
 impl<P: Atomic> GenericLocalCounter<P> {
     fn new(counter: GenericCounter<P>) -> Self {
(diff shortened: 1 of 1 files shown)


src/counter.rs
```diff
@@ -571,38 +571,4 @@ mod tests {
...
-    }
-
-    #[cfg(debug_assertions)]
-    #[test]
-    #[should_panic(expected = "assertion failed")]
-    fn test_local_counter_negative_inc() {
-        let counter = Counter::new("foo", "bar").unwrap();
-        let local = counter.local();
-        local.inc_by(-42.0);
-    }
```
breeswish, line 590: I think only tests for `IntCounter` need to be removed. `Counter` accepts floats which can be negative.

src/counter.rs
```diff
@@ -25,7 +25,7 @@ pub type Counter = GenericCounter<AtomicF64>;
 
 /// The integer version of [`Counter`]. Provides better performance if metric values
 /// are all integers.
```
mxinden, line 27:
```suggestion
/// are all positive integers (natural numbers).
```

src/counter.rs
```diff
@@ -142,7 +142,7 @@ pub type CounterVec = GenericCounterVec<AtomicF64>;
 
 /// The integer version of [`CounterVec`]. Provides better performance if metric
 /// values are all integers.
```
mxinden, line 144:
```suggestion
/// are all positive integers (natural numbers).
```

src/counter.rs
```diff
@@ -186,7 +186,7 @@ pub type LocalCounter = GenericLocalCounter<AtomicF64>;
 
 /// The integer version of [`LocalCounter`]. Provides better performance
 /// if metric values are all integers.
```
mxinden, line 188:
```suggestion
/// are all positive integers (natural numbers).
```

src/counter.rs
```diff
@@ -272,7 +272,7 @@ pub type LocalCounterVec = GenericLocalCounterVec<AtomicF64>;
 
 /// The integer version of [`LocalCounterVec`].
 /// Provides better performance if metric values are all integers.
```
mxinden, line 274:
```suggestion
/// Provides better performance if metric values are all positive
/// integers (natural numbers).
```

mxinden (approved):
I am in favor of this change. Thanks.

I find the name **Int**Counter a bit confusing, as my intuition would include negative numbers as well, which (rightly so) doesn't make any sense for a Prometheus counter. I am failing to come up with a better name.

While already enforced through the type system, I suggest to make the following changes to the type signature comments to emphasize the fact that one can only to pass natural numbers.

src/counter.rs
Folyd: Thanks for reviewing. I totally agree.

tim-seoss:
It might be useful to also document (not sure where's best, maybe along with the "better performance" notes?), that these integer counters will wrap with `inc` and `inc_by` (and that this is acceptable, so long as it doesn't wrap more than once between samples).

Not sure if that's in scope for this PR tho'.

mxinden:
> It might be useful to also document (not sure where's best, maybe along with the "better performance" notes?), that these integer counters will wrap with inc and inc_by (and that this is acceptable, so long as it doesn't wrap more than once between samples).

I am in favor of this, but as you said, I think it is out of scope for this pull request. @tim-seoss would you mind opening up a new pull request doing the suggested changes?

mxinden:
Unless there are any objections I will merge this pull request later today.

breeswish (approved):
Thank you!

PR #366, opened 2020-11-06
benches: move to criterion-rs and GitHub actions

lucab:
This moves all benches to criterion-rs, in order to allow running
benchmarks on a stable toolchain.
This also moves benchmarking and linting jobs to a pinned toolchain,
under GitHub actions.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,45 @@
+---
+name: Rust
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+env:
+  CARGO_TERM_COLOR: always
+  # Pinned toolchain for linting and benchmarks
+  ACTIONS_LINTS_TOOLCHAIN: 1.47.0
+
+jobs:
+  linting:
+    name: "Lints, pinned toolchain"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ env['ACTIONS_LINTS_TOOLCHAIN']  }}
+          default: true
+          components: rustfmt, clippy
+      - name: cargo fmt (check)
+        run: cargo fmt --all -- --check -l
+      - name: cargo clippy
+        run: cargo clippy --all -- -D clippy
+  criterion:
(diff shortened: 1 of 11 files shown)


lucab:
This was green on my fork: https://github.com/lucab/rust-prometheus/actions/runs/349778558

.github/workflows/rust.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ env['ACTIONS_LINTS_TOOLCHAIN']  }}
+          default: true
+      - name: cargo bench
+        run: cargo bench
```
mxinden, line 43:
```suggestion
        run: cargo bench --all-features
```

Would this work as well?

.travis.yml
```diff
@@ -46,6 +40,5 @@ script:
         cargo test -p prometheus-static-metric --no-default-features --features="$FEATURES"
 
         # Run benchmarks.
-        cargo bench --no-default-features --features="$FEATURES"
         cargo bench -p prometheus-static-metric --no-default-features --features="$FEATURES"
```
mxinden, line 43: I am a bit reluctant to only partially move to criterion. Is there something blocking us from using criterion in prometheus-static-metric as well?

mxinden (commented):
I am in favor of this. Thanks for doing the change.

.travis.yml
breeswish: Maybe because `prometheus-static-metric` itself is something that requires nightly>

.github/workflows/rust.yml
lucab: Not really because of the `gen` and `nightly` features. But we aren't really benchmarking anything that requires extra features, so I think this is fine at least for now.

.travis.yml
lucab: No sorry, I just ran out of time on Friday and preferred to split the move into smaller bits. I saw today that there wasn't too much left to converted, so I amended this PR now to complete the move.

lucab:
This should be ready for another round of reviews.

Checks will not show up here until the workflow is merged first, but they were green on my fork: https://github.com/lucab/rust-prometheus/actions/runs/353547239

mxinden (approved):

PR #372, opened 2020-12-04
static-metric/Cargo.toml: Remove include section

mxinden:
Releasing static-metric fails with:

```
Caused by:
  failed to parse manifest at
`/xxx/tikv/rust-prometheus/target/package/prometheus-static-metric-0.5.0/Cargo.toml`

Caused by:
  can't find `benches` bench, specify bench.path
```

This is due to the `benches` folder not being included in the final
release tarball. With this commit the folder is included and thus
releasing prometheus-static-metric would succeed.

Off the top of my head this is the right way to go. Open for alternative
suggestions.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -28,6 +28,8 @@ jobs:
         run: cargo fmt --all -- --check -l
       - name: cargo clippy
         run: cargo clippy --all -- -D clippy
+      - name: cargo package
+        run : cargo package && cargo package --manifest-path static-metric/Cargo.toml
   criterion:
     name: "Benchmarks (criterion)"
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


lucab:
Do we need the `include` list at all? It seems to be missing other paths too (e.g. `tests/`) and contain bogus entries (`LICENSE`).

Whatever outcome, CI can also validate the fix here with `cargo package --manifest-path static-metric/Cargo.toml` as an additional workflow step.

mxinden:
I followed both suggestions by @lucab. (1) introducing a new check to CI (see failing test on https://github.com/tikv/rust-prometheus/pull/372/commits/e76706dc027a41e832d141507efab5ac58d7bd38) and (2) remove the `includes` section. Thanks for the suggestions.

@breeswish as far as I can tell you introduced the `includes` section. Are you fine with this pull request removing it? Anything we are missing?

lucab (approved):

breeswish (approved):

mxinden:
Release [tagged](https://github.com/tikv/rust-prometheus/tree/prometheus-static-metric-v0.5.0) and [pushed](https://crates.io/crates/prometheus-static-metric/0.5.0).

PR #381, opened 2021-01-07
src/: Replace deprecated compare_and_swap with compare_exchange(_weak)

mxinden:
`compare_and_swap` is deprecated in favor of `compare_exchange` and
`compare_exchange_weak`. This commit replaces the former with either of
the two latter options.

Also see [migration guide](https://doc.rust-lang.org/beta/std/sync/atomic/struct.AtomicU64.html#migrating-to-compare_exchange-and-compare_exchange_weak).

I welcome anyone to review this change. Given that this is a rather subtle change (especially with memory ordering) possibly resulting in a much greater disaster, please don't merge this just yet. I would like to review this myself once more with a fresh mind.

`

### DIFF
--- a/src/atomic64.rs
+++ b/src/atomic64.rs
@@ -217,7 +217,7 @@ impl AtomicU64 {
     /// and containing the previous value.
     ///
     /// See [`StdAtomicU64`] for details.
-    pub fn compare_exchange_weak(
+    pub(crate) fn compare_exchange_weak(
         &self,
         current: u64,
         new: u64,
--- a/src/atomic64.rs
+++ b/src/atomic64.rs
@@ -110,10 +110,13 @@ impl Atomic for AtomicF64 {
         loop {
             let current = self.inner.load(Ordering::Acquire);
             let new = u64_to_f64(current) + delta;
-            let swapped = self
-                .inner
-                .compare_and_swap(current, f64_to_u64(new), Ordering::Release);
-            if swapped == current {
+            let result = self.inner.compare_exchange_weak(
+                current,
+                f64_to_u64(new),
+                Ordering::Release,
+                Ordering::Relaxed,
+            );
+            if result.is_ok() {
                 return;
             }
         }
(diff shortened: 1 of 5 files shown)


src/atomic64.rs
```diff
@@ -205,9 +208,24 @@ impl Atomic for AtomicU64 {
 }
 
 impl AtomicU64 {
-    /// Get the value with the provided memory ordering.
-    pub fn compare_and_swap(&self, current: u64, new: u64, ordering: Ordering) -> u64 {
```
lucab, line 209: Self-note: this is an API break.

src/atomic64.rs
```diff
@@ -205,9 +208,24 @@ impl Atomic for AtomicU64 {
...
+    /// Stores a value into the atomic integer if the current value is the same
+    /// as the current value.
+    ///
+    /// This function is allowed to spuriously fail even when the comparison
+    /// succeeds, which can result in more efficient code on some platforms. The
+    /// return value is a result indicating whether the new value was written
+    /// and containing the previous value.
+    ///
+    /// See [`StdAtomicU64`] for details.
+    pub fn compare_exchange_weak(
```
lucab, line 220: On this one, I think you'd want to expose the non-weak version.

lucab (commented):
My understanding is that the non-weak version is the drop-in replacement, while the weak one may be used inside hot-loops as an optimization. Thus, I mostly agree with your choices, except for the public method on `AtomicU64`.

src/atomic64.rs
mxinden: `AtomicU64::compare_exchange_weak` is only used in `HistogramCore::proto`. As `HistogramCore::proto` is trying to compare and set a value in a loop it should suffice to use `compare_exchange_weak` instead of `compare_exchange`.

If I am not mistaken `AtomicU64` is not accessible when used in a metric like `GenericCounter`, thus I don't think we need to expose `compare_exchange` as well in order to maintain backward-compatibility.

Does the above make sense @lucab?

lucab: My line of reasoning was that it is a public type+method, so if people are using this for whatever reason, then they'd want to have a 1-to-1 replacement when upgrading to the new version.

As it was just recently introduced by your atomic-histogram PR, likely nobody is yet using this, thus I agree it is fine to mark this a `pub(crate)` and keep the weak version.

mxinden:
Made `compare_exchange_weak` `pub(crate)` and cherry-picked https://github.com/tikv/rust-prometheus/pull/380/. @lucab can you approve in case this looks good to you?

lucab (approved):

PR #388, opened 2021-02-09
Feat/254 int histogram

hdost:


src/lib.rs
```diff
@@ -217,9 +217,11 @@ pub use self::encoder::TextEncoder;
 pub use self::encoder::{PROTOBUF_FORMAT, TEXT_FORMAT};
 pub use self::errors::{Error, Result};
 pub use self::gauge::{Gauge, GaugeVec, IntGauge, IntGaugeVec};
+#[deprecated(since = "0.12.0", note = "Please use DEFAULT_FLOAT_BUCKETS")]
 pub use self::histogram::DEFAULT_BUCKETS;
```
breeswish, line 221: Considering that Histogram is called Histogram instead of FloatHistogram, I suggest to keep `DEFAULT_BUCKETS` and not being `DEFAULT_FLOAT_BUCKETS`. What do you think?

src/histogram.rs
```diff
@@ -41,9 +48,34 @@ fn check_bucket_label(label: &str) -> Result<()> {
     Ok(())
 }
 
-fn check_and_adjust_buckets(mut buckets: Vec<f64>) -> Result<Vec<f64>> {
+fn check_and_adjust_int_buckets(mut buckets: Vec<u64>) -> Result<Vec<u64>> {
```
breeswish, line 51: Maybe we can use generics to avoid writing code twice.

src/histogram.rs
```diff
@@ -386,18 +560,55 @@ impl HistogramCore {
...
+        // the hot and cold shards.
+        let _guard = self.collect_lock.lock().expect("Lock poisoned");
+
+        let (shard_index, _count) = self.shard_and_count.get();
+        self.shards[shard_index as usize].sum.get()
+    }
+
+}
+
+impl IntHistogramCore {
```
breeswish, line 574: We'd better avoid duplicating these codes. You can try with generics!

src/histogram.rs
hdost: From what i can tell this would require a bit of a re-architecture. I will take a closer look this may be a breaking change and see what i can do.

PR #390, opened 2021-02-19
Allow trailing commas on macros

HyeonuPark:
This PR allows to append optional comma after the last argument of those opts!/register! like macros.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -36,6 +36,17 @@ macro_rules! labels {
     };
 }
 
+#[test]
+fn test_labels_without_trailing_comma() {
+    let labels = labels! {
+        "test" => "hello",
+        "foo" => "bar"
+    };
+    assert_eq!(labels.len(), 2);
+    assert!(labels.get("test").is_some());
+    assert_eq!(*(labels.get("test").unwrap()), "hello");
+}
+
 /// Create an [`Opts`].
 ///
 /// # Examples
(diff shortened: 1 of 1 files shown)


HyeonuPark:
I guess all I needs to do for the DCO issue is to append the `Signed-off-by:` line to the commit msg. Maybe it was a mistake to make the change directly on the GH web editor, which makes it nontrivial to simply change the commit msg :P

lucab:
@HyeonuPark thanks for the PR! Indeed, your commit should be signed off to pass CI checks. Travis failure is due to https://github.com/tikv/rust-prometheus/pull/391 and should go away if you rebase to latest master.

lucab (approved):

breeswish (approved):
Thanks a lot! Could you also supply some unit tests to verify that this syntax pass the compiling? In this way we can avoid some future mistakes :) For example, you can just write a test importing the macro and then use these macros with trailing commas.

lucab:
@HyeonuPark ping, do you plan to come back to this soon-ish?

HyeonuPark:
Sorry for the super late reply! If it's still OK, I'll supplement it within this week.

HyeonuPark:
Shameless copy-pastes from doc comments with additional comma per invocation. :) I can't say anymore it's a tiny change w.r.t. the LoC.

HyeonuPark:
Applied cargo fmt on it.

lucab:
Thanks a lot, LGTM!

PR #391, opened 2021-02-19
desc: fix format string for panic

lucab:
This fixes a compile error on current beta toolchain, which started
enforcing that panic messages must be string literals (with format
specifiers).


lucab:
Full error is:
```
error: panic message is not a string literal
   --> src/desc.rs:260:33
    |
260 |                 other => panic!(other),
    |                                 ^^^^^
    |
    = note: `-D non-fmt-panic` implied by `-D warnings`
```

mxinden (approved):

breeswish (approved):

Issue #393, opened 2021-02-24
Exemplar support


clux:
It would be awesome to have the ability to attach trace ids on histograms and counters as per the [OpenMetics specifictaion](https://github.com/OpenObservability/OpenMetrics/blob/master/specification/OpenMetrics.md#exemplars-1) so that we can link metric data points to traces (like in [grafana tempo](https://grafana.com/oss/tempo/)).

**Proposed Solution**
Something similar to what the [golang client has implemented](https://github.com/prometheus/client_golang/pull/706/files), so that the encoders can include the exemplars as in OM:

```
# TYPE foo histogram
foo_bucket{le="0.01"} 0
foo_bucket{le="0.1"} 8 # {} 0.054
foo_bucket{le="1"} 11 # {trace_id="KOO5S4vxi0o"} 0.67
foo_bucket{le="10"} 17 # {trace_id="oHg5SJYRHA0"} 9.8 1520879607.789
foo_bucket{le="+Inf"} 17
foo_count 17
foo_sum 324789.3
foo_created  1520430000.123
```

**Describe alternatives you've considered**
Using a paid solution like HoneyComb.

**Additional context**
huge thanks for this providing library, it gets huge mileage

hsyed:
I could work on this. I contributed the Openmetrics/exemplars support for Prometheus-Dotnet. 

I am learning Rust, this would be good fun.

I haven't looked at the codebase closely yet, but I feel the key points of the work are:

1. Add new collection methods to the instruments for capturing exemplars.
2. Update the data model to carry examplers with the samples.
3. Introduce a new Encoder for OpenMetrics.
4. Add a helper function to negotiate the correct encoder given the http header string.

What do the maintainers think ? (cc: @lucab)

PR #395, opened 2021-02-26
Add exemplar support

clux:
This is a quick go at adding exemplar support from #393.

Is this something you would be willing to accept if it was cleaned up?

Have added them to counters atm, but looks like there are a lot of layers between the Counter and the encoding bits so haven't gotten that far. If this is welcome, then any pointers on how to best propagate the data to the encoders would be appreciated.


clux:
Well, I've done the propagation..__one way__. It was hard to retrofit the whole cold/hot shard stuff with an extra optional struct which I don't know how to do atomic style for.

Also not cared for protobuf beyond what was necessary.

But can now do:

```rust
    let mut exemplar_labels = HashMap::new();
    exemplar_labels.insert("traceID".into(), trace_id);
    let ex = Exemplar::new_with_labels(duration, exemplar_labels);
    reconcile_duration_histogram
        .with_label_values(&[])
        .observe_with_exemplar(duration, ex);
```

and text encoder outputs:

```
# HELP foo_controller_reconcile_duration_seconds The duration of reconcile to complete in seconds
# TYPE foo_controller_reconcile_duration_seconds histogram
foo_controller_reconcile_duration_seconds_bucket{le="0.01"} 7 # {traceID="53b9b11e6aa06934c5f1bcedc772130d"}  37
foo_controller_reconcile_duration_seconds_bucket{le="0.1"} 8
foo_controller_reconcile_duration_seconds_bucket{le="0.25"} 8
```

which means the date probably isn't working, but it looks like it conforms to the spec at least.

again, comments would be helpful. i feel like i am just hacking around in the proto_model and histogram files.

clux:
have had to subvert the histogramcore sharding a bit, wasn't sure it made any sense to do a hot/cold path for exemplars when you have to sync write the whole struct, but maybe i am wrong.

for now it's sitting straight as a vector of optional exemplars (same length as bucket len) and writes to them are sync, which probably makes the existing sync mechanism useless. not sure how to make this better. could feature flag exemplars?

wperron:
Is there anything left to do on this PR? Anything I can do to help?

clux:
Well, I guess the main thing is that no matter how much I wrangle prometheus, it does not accept my output, even though it looks like it should conform to the spec. I have a demo app that is outputting:

```
foo_controller_reconcile_duration_seconds_bucket{le="0.01"} 1 # {trace_id="6c1011d34f17c4ca1177b40ddfb5184b"} 0.006
foo_controller_reconcile_duration_seconds_bucket{le="0.1"} 1
foo_controller_reconcile_duration_seconds_bucket{le="0.25"} 1
```
in its /metrics

but prometheus fails to parse it with a:

```
expected timestamp or new record, got "MNAME"
```

I have a demo implementation in https://github.com/clux/controller-rs/pull/12, and it's definitely the exemplar output that is failing, so not really sure what to do :-(

bobrik:
Exemplars must have the timestamp after the example value, just like the error says.

clux:
@bobrik : The error is asking for a histogram timestamp or a new record, not an exemplar timestamp (which is optional by the spec). These two are both valid:

```sh
$ echo 'foo_controller_reconcile_duration_seconds_bucket{le="0.01"} 1' | promtool check metrics
foo_controller_reconcile_duration_seconds_bucket no help text
$ echo 'foo_controller_reconcile_duration_seconds_bucket{le="0.01"} 1 1623574027' | promtool check metrics
foo_controller_reconcile_duration_seconds_bucket no help text
```

But none of these are:

```sh
$ echo 'foo_controller_reconcile_duration_seconds_bucket{le="0.01"} 1 1623574027 # {trace_id="69d991d2ef39b15e71e2f42eb02a9b55"} 0.006"' | promtool check metrics
error while linting: text format parsing error in line 1: spurious string after timestamp: " # {trace_id=\"69d991d2ef39b15e71e2f42eb02a9b55\"} 0.006\""
$ echo 'foo_controller_reconcile_duration_seconds_bucket{le="0.01"} 1 1623574027 # {trace_id="69d991d2ef39b15e71e2f42eb02a9b55"} 0.006 1623574027"' | promtool check metrics
error while linting: text format parsing error in line 1: spurious string after timestamp: " # {trace_id=\"69d991d2ef39b15e71e2f42eb02a9b55\"} 0.006 1623574027\""
$ echo 'foo_controller_reconcile_duration_seconds_bucket{le="0.01"} 1 1623574027 # ' | promtool check metrics
error while linting: text format parsing error in line 1: spurious string after timestamp: " # "
```

and without both optional timestamps (orig error):

```sh
$ echo 'foo_controller_reconcile_duration_seconds_bucket{le="0.01"} 1 # {trace_id="69d991d2ef39b15e71e2f42eb02a9b55"} 0.006"' | promtool check metrics
error while linting: text format parsing error in line 1: expected integer as timestamp, got "#"
```

roidelapluie:
Prometheus has 2 metrics format: Prometheus text format, and OpenMetrics.

Only OpenMetrics supports examplars. There are a few other differences between the Text Format and OpenMetrics.

clux:
Thanks a lot @roidelapluie ! That solves the mystery.

I can get this client to work with openmetrics format, but it required a different exposition ending (`# EOF\n`), and forces a user requirement on the new content-type header, and possibly a different timestamp format (which I started tackling in here, but haven't changed to/verified anywhere).

So while this does seem parseable by prometheus if we skip optional exemplar timestamps and include the header, I am starting to fear that this PR might be a big one to merge if it requires changing the entire content format :(

roidelapluie:
We are not evolving the text format anymore, we want to promote OpenMetrics. If you plan for adding full openmetrics support for Rust, do not hesitate to reach for help. There are a few differences between both formats, and the library ideally would handle both (via negotiation in the headers).

PR #397, opened 2021-03-05
prometheus: release 0.12.0

lucab:
## 0.12.0

 - Improvement: Fix format string in panic!() calls (#391)

 - Improvement: Replace regex with memchr (#385)

 - Improvement: Update reqwest requirement from ^0.10 to ^0.11 (#379)

Closes: https://github.com/tikv/rust-prometheus/issues/394

### DIFF
--- a/static-metric/Cargo.toml
+++ b/static-metric/Cargo.toml
@@ -19,7 +19,7 @@ lazy_static = "1.4"
 
 [dev-dependencies]
 criterion = "0.3"
-prometheus = { version = "0.11.0", path = "../" }
+prometheus = { path = "../" }
 
 [features]
 default = []
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.12.0
+
+ - Improvement: Fix format string in panic!() calls (#391)
+
+ - Improvement: Replace regex with memchr (#385)
+
+ - Improvement: Update reqwest requirement from ^0.10 to ^0.11 (#379)
+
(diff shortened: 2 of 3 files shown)


static-metric/Cargo.toml
```diff
@@ -19,7 +19,7 @@ lazy_static = "1.4"
 
 [dev-dependencies]
 criterion = "0.3"
-prometheus = { version = "0.11.0", path = "../" }
+prometheus = { version = "0.12.0", path = "../" }
```
lucab, line 22: I guess I'll have to drop the `cargo package static-metric` test in CI otherwise we end up in a situation which is impossible to satisfy.

mxinden, line 22:
```suggestion
prometheus = { path = "../" }
```

Is the version needed in the first place? In https://github.com/libp2p/rust-libp2p/ we use relative paths only to specify dev-dependencies in the same repository to get around the circular dependency issue.

mxinden (approved):
Cool, thanks for doing the release.

static-metric/Cargo.toml
lucab: As far as I know it is needed otherwise `cargo` won't let you publish/upload it. But maybe some changed or my knowledge is incomplete.

lucab: Ah nevermind there is now a special case for dev-dependencies documented at https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations. So yes I think it can be dropped.

PR #400, opened 2021-05-07
Implement methods on `Registry` for managing prefix and common labels

alexforster:
This allows users to set the registry's prefix or common labels when using the default registry.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -263,6 +263,67 @@ impl Registry {
         Ok(reg)
     }
 
+    /// `prefix` returns a copy of the prefix for the registry, if any.
+    pub fn prefix(&self) -> Option<String> {
+        self.r.read().prefix.clone()
+    }
+
+    /// `set_prefix` sets the prefix for the registry, and returns the previous
+    /// prefix, if any. If `prefix` is empty, the function returns an error.
+    pub fn set_prefix(&self, prefix: &str) -> Result<Option<String>> {
+        if prefix.is_empty() {
+            return Err(Error::Msg("empty prefix namespace".to_string()));
+        }
+
+        let mut core = self.r.write();
+        let result = core.prefix.take();
+        core.prefix = Some(prefix.to_string());
+        Ok(result)
+    }
+
+    /// `clear_prefix` removes the prefix from the registry and returns it to
+    /// the caller, if a prefix was set.
+    pub fn clear_prefix(&self) -> Option<String> {
+        let mut core = self.r.write();
+        core.prefix.take()
+    }
+
+    /// `labels` returns a copy of the common labels for the registry, if any.
+    pub fn labels(&self) -> Option<HashMap<String, String>> {
+        self.r.read().labels.clone()
+    }
+
+    /// `insert_label` adds or updates a label in the set of common labels for
+    /// the registry, and returns the previous value of that label, if any.
+    /// If either `key` or `value` are empty, the function returns an error.
+    pub fn insert_label(&self, key: &str, value: &str) -> Result<Option<String>> {
+        if key.is_empty() {
+            return Err(Error::Msg("empty label key".to_string()));
+        }
+        if value.is_empty() {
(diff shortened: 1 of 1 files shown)


src/registry.rs
```diff
@@ -263,6 +263,67 @@ impl Registry {
...
+
+    /// `clear_prefix` removes the prefix from the registry and returns it to
+    /// the caller, if a prefix was set.
+    pub fn clear_prefix(&self) -> Option<String> {
+        let mut core = self.r.write();
+        core.prefix.take()
+    }
+
+    /// `labels` returns a copy of the common labels for the registry, if any.
+    pub fn labels(&self) -> Option<HashMap<String, String>> {
```
lucab, line 292: Here and in `prefix()` above, can we return a reference instead (and let the consumers clone if they want)?

lucab:
Thanks for the PR! In order to merge, this will need a rebase and a `Signed-off-by` on each commit.

I think we'd need a bit more context behind this. I do like the getters (minus the clone part), but I'm not thrilled by the set/clean logic. I won't expect people to use the registry that way, and the fact that those methods internally lock will likely become a bad performance surprise for many people.
I'd prefer sticking to the "set options on build" pattern and only keep the getters here.

polachok:
I think the problem this PR is inspired by is that it's `register_*` macros use default registry and creating metrics with non-default registries is very verbose. At least that's what I found in my use and this PR helps a lot, I can just add labels to the default registry and use macros as usual.

Disclaimer: I'm not the author, they may have another use case.

lucab:
@polachok thanks, that was useful feedback! Next release should come with https://github.com/tikv/rust-prometheus/pull/396 which should hopefully improve that UX aspect, but taking a different path from this PR.

That said, I'm still intimately not in favor of doing this kind of registry mutation at runtime. It's a bazooka with side-effects at distance, which can easily turn into a very large footgun.

alexforster:
> I think we'd need a bit more context behind this.

Right, I think based on your feedback that I wasn't clear on the use-case. Specifically, this is intended to make it possible to manipulate the set of default labels on [the default registry](https://github.com/tikv/rust-prometheus/blob/f8d3bb2a0eed2c9bdeb87e9d81a47ea36e627fbc/src/registry.rs#L373). I do not believe that this adds any useful functionality for custom registries. However, I think custom registries are probably a very niche use-case anyway:

- I've never had a need to create more than one registry in any codebase. I can't even think of a hypothetical scenario where I'd want to.
- Creating a custom registry and passing it around everywhere is a huge chore. Metrics are everywhere in a typical codebase, so the registry object ends up infecting every corner of a system.

Because of this, I always use the default registry, and I bet 99% of your users do too.

> I do like the getters (minus the clone part), but I'm not thrilled by the set/clean logic.

Keeping getters and removing setters doesn't add any meaningful functionality imo. The functionality I'm trying to add here is the ability to set default labels on the default registry.

> I won't expect people to use the registry that way, and the fact that those methods internally lock will likely become a bad performance surprise for many people. I'd prefer sticking to the "set options on build" pattern and only keep the getters here.

Again, this is only for default labels on a Registry. Presumably default labels would only need to be configured once, not continuously.

alexforster:
I've rebased and added a `Signed-off-by`

adeschamps:
I thought I was going to need this feature, and was about to use this branch via a `[patch.crates-io]` entry. I wanted to expose a build version and other information that's static over the lifetime of the process. I ended up exposing it via a single gauge metric, as described in the article [Exposing the software version to Prometheus](https://www.robustperception.io/exposing-the-software-version-to-prometheus/) instead (which I discovered via a recent blog post from the `autometrics` crate).

I do have a use case for multiple registries though -- we have an application that we deliver to users, and also operate as a hosted service. Some metrics measure implementation details or unstable features, so we initially add them to the default registry (which is only exposed in internal builds) and later switch them to a different registry when we're sure that they're useful for users and unlikely to be changed.

PR #401, opened 2021-05-17
Export thread count from `process_collector`

free:
This simply retrieves and exports the value of the `num_threads` field from the
already populated `procfs::process::Stat` that CPU, memory and start time are
sourced.

### DIFF
--- a/src/process_collector.rs
+++ b/src/process_collector.rs
@@ -17,12 +17,12 @@ use crate::proto;
 /// The `pid_t` data type represents process IDs.
 pub use libc::pid_t;
 
-/// Six metrics per ProcessCollector.
-const METRICS_NUMBER: usize = 6;
+/// Seven metrics per ProcessCollector.
+const METRICS_NUMBER: usize = 7;
 
-/// A collector which exports the current state of
-/// process metrics including cpu, memory and file descriptor usage as well as
-/// the process start time for the given process id.
+/// A collector which exports the current state of process metrics including
+/// CPU, memory and file descriptor usage, thread count, as well as the process
+/// start time for the given process id.
 #[derive(Debug)]
 pub struct ProcessCollector {
     pid: pid_t,
@@ -33,6 +33,7 @@ pub struct ProcessCollector {
     vsize: Gauge,
     rss: Gauge,
     start_time: Gauge,
+    threads: Gauge,
 }
 
 impl ProcessCollector {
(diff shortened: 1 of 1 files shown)


free:
Gentle nudge.

lucab:
Thanks for the PR. Please note however that the ProcessCollector implements https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics. I'm thus wary of adding more arbitrary stuff directly here.

I'd be happy to see the "Number of OS threads" metrics standardized upstream, and once that is published we can promptly merge this.

free:
> I'd be happy to see the "Number of OS threads" metrics standardized upstream, and once that is published we can promptly merge this.

Is anything like that being worked on? On the one hand the Golang library [has `go_threads`](https://github.com/prometheus/client_golang/blob/b89620c4916814c2960f16255c36d1b381cda9e7/prometheus/go_collector.go#L49) and Java [has `jvm_threads_current`](https://github.com/prometheus/client_java/blob/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java#L25-L31) (and I'm sure other language libraries have their own).

On the other hand, the only other way of getting this metric (other than everyone rolling their own implementation, which is what we ended up doing) is to enable the (by default disabled) [`processes` collector](https://github.com/prometheus/node_exporter#disabled-by-default) of `node-exporter` (above which there's a warning that the collector may be disabled due to "prolonged runtime" or "significant resource usage"). While adding the metric here boils down to simply exporting a value that has already been retrieved from the OS.

I'm not going to insist that this be merged, but the fact that `process_threads` is not a standard process metric is not a good reason against exporting it regardless.

lucab:
I'm not aware of any such discussion, so I started it at https://github.com/prometheus/docs/pull/1964. As you mentioned, it is a common metrics across libraries so it makes sense to have it uniformed. Plus, the `process_` prefix is reserved and I think that having to name this `rust_process_threads` does not make much sense here.

src/process_collector.rs
```diff
@@ -100,6 +101,12 @@ impl ProcessCollector {
         .unwrap();
         descs.extend(start_time.desc().into_iter().cloned());
 
+        let threads = Gauge::with_opts(
+            Opts::new("process_threads", "Number of OS threads.").namespace(namespace.clone()),
```
lucab, line 105: `Number of OS threads in the process.`

lucab:
@free sorry for putting this through the long path. The upstream doc change has been merged, so this can land too. It only needs a rebase and a minor rewording.

src/process_collector.rs
free: Done,

free:
Phew, took a bit to get the DCO right.

> sorry for putting this through the long path.

No worries, thanks for following through. I put together a polyfill locally in the meantime, so no problem.

lucab (approved):

PR #402, opened 2021-07-08
Add convenience TextEncoder functions to encode directly to string

matklad:
I saw that we do

```rust
        let mut buffer = vec![];
        let encoder = TextEncoder::new();
        encoder.encode(&prometheus::gather(), &mut buffer).unwrap();

        String::from_utf8(buffer).unwrap()
```

a bunch of times in [near](github.com/near/nearcore) and figured that the library can probably guaranty utf8 validity and remove the need for one unwrap here. 

This turned out to be slightly more involved then I've thought, as there's no standard way to explain that you'll only write utf8 to a `Write`, but it wasn't to hard either. 

As a bonus, `TextEncoder` impl now is statically guaranteed to produce only valid utf8

### DIFF
--- a/src/encoder/text.rs
+++ b/src/encoder/text.rs
@@ -1,8 +1,7 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
 use std::borrow::Cow;
-use std::io::Write;
-use std::mem;
+use std::io::{self, Write};
 
 use crate::errors::Result;
 use crate::histogram::BUCKET_LABEL;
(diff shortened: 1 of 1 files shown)


src/encoder/text.rs
```diff
@@ -25,10 +26,27 @@ impl TextEncoder {
...
+    /// This is a convenience wrapper around `<TextEncoder as Encoder>::encode`.
+    pub fn encode_utf8(&self, metric_families: &[MetricFamily], buf: &mut String) -> Result<()> {
+        // Note: it's important to *not* re-validate ut8 validity for the
+        // entirety of `buf`. Otherwise, repeatedly appending metrics to the
+        // same `buf` will lead to quadratic behavior. That's why we use
+        // `Utf8Writer` abstraction to skip the validation.
+        Utf8Writer::write_to_string(buf, |writer| self.encode_impl(metric_families, writer))?;
+        Ok(())
+    }
+    /// Converts metrics to `String.
```
lucab, line 40: Missing trailing ``` ` ```.

src/encoder/text.rs
```diff
@@ -25,10 +26,27 @@ impl TextEncoder {
     pub fn new() -> TextEncoder {
         TextEncoder
     }
-}
+    /// Appends metrics to a given `String` buffer.
+    ///
+    /// This is a convenience wrapper around `<TextEncoder as Encoder>::encode`.
+    pub fn encode_utf8(&self, metric_families: &[MetricFamily], buf: &mut String) -> Result<()> {
+        // Note: it's important to *not* re-validate ut8 validity for the
```
lucab, line 33: Typo: "UTF-8 validity"

lucab:
Thanks for PR. Looks great overall, I've left a couple of minor comments in reviews.
It looks like our CI is having a bad day; I'll try to have a look at fixing it as soon as I have a free timeslot, then this PR may need to be rebased to see a green run.

matklad:
Sure, no problem rebasing!

FYI, I see the "1 workflow awaiting approval" banner, if you want to disable for new contributors, you can check this box:

image

matklad:
Figured out a much better solution -- there's no need for unsafety here.

lucab:
Ok, https://github.com/tikv/rust-prometheus/pull/403 is merged so rebasing this PR now should make the CI happy.

lucab:
@zhouqiang-cl @BusyJay @breeswish whoever of you has access to do that, can you please tweak the workflow permissions as shown in https://github.com/tikv/rust-prometheus/pull/402#issuecomment-877115118 so that we don't have to re-approve it on every push?

BusyJay:
As github actions resource are bound to the organization account, so it might not be a good idea to just change it for a repository. Maybe you can send an issue to tikv/community or a thread at https://internals.tidb.io/, let's discuss it more!

matklad:
@lucab ptal :)

lucab (approved):
LGTM, thanks!

PR #405, opened 2021-07-25
Add PullingGauge gauge type

JanBerktold:
First of all, thank you for this crate! We've been using it in production for a bit and are super happy.

As part of this, we've written a few additional metric types that allow us to instrument our app in an easier fashion. This PR attempts to merge the first of these back to the upstream project. If there is interest, then we'd be happy to send more PRs with other helper types (i.e. we have a MaximumOverIntervalGauge that allows monitoring of quickly changing values).

The metric type that this PR is adding is the **PullingGauge**. It is useful in scenarios where we want to expose a particular value but it is infeasible to keep the metric up to date all of the time (i.e. it changes a lot in hot code paths). The most common example for us internally is to expose the size of a _Vector_ or similiar collections. In those cases, we create one of these and pass a `move || vector.len() as f64` as the lambda.

I've more or less just lifted this out of our internal codebase -- please let me know in case of any changes required to make it fit better in this crate!

### DIFF
--- a/src/pulling_gauge.rs
+++ b/src/pulling_gauge.rs
@@ -1,0 +1,101 @@
+use std::{collections::HashMap, fmt, sync::Arc};
+
+use crate::{
+    core::Collector,
+    proto::{Gauge, Metric, MetricFamily, MetricType},
+};
+use protobuf::RepeatedField;
+
+/// A [Gauge] that returns the value from a provided function on every collect run.
+///
+/// This metric is the equivalant of Go's
+/// <https://pkg.go.dev/github.com/prometheus/client_golang@v1.11.0/prometheus#GaugeFunc>
+///
+/// # Examples
+/// ```
+/// # use prometheus::{Registry, PullingGauge};
+/// # // We are stubbing out std::thread::available_parallelism since it's not available in the
+/// # // oldest Rust version that we support.
+/// # fn available_parallelism() -> f64 { 0.0 }
+///
+/// let registry = Registry::new();
+/// let gauge = PullingGauge::new(
+///     "available_parallelism",
+///     "The available parallelism, usually the numbers of logical cores.",
+///     Box::new(|| available_parallelism())
+/// ).unwrap();
+/// registry.register(Box::new(gauge));
+/// ```
+#[derive(Clone)]
+pub struct PullingGauge {
+    desc: crate::core::Desc,
+    value: Arc<Box<dyn Fn() -> f64 + Send + Sync>>,
+}
+
+impl fmt::Debug for PullingGauge {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("PullingGauge")
+            .field("desc", &self.desc)
+            .field("value", &"<opaque>")
+            .finish()
+    }
+}
+
+impl PullingGauge {
+    /// Create a new [`PullingGauge`].
+    pub fn new<S1: Into<String>, S2: Into<String>>(
+        name: S1,
+        help: S2,
+        value: Box<dyn Fn() -> f64 + Send + Sync>,
+    ) -> crate::Result<Self> {
+        Ok(PullingGauge {
+            value: Arc::new(value),
+            desc: crate::core::Desc::new(name.into(), help.into(), Vec::new(), HashMap::new())?,
+        })
+    }
+
+    fn metric(&self) -> Metric {
+        let mut gauge = Gauge::default();
+        let getter = &self.value;
+        gauge.set_value(getter());
+
+        let mut metric = Metric::default();
+        metric.set_gauge(gauge);
+
+        metric
+    }
+}
+
+impl Collector for PullingGauge {
+    fn desc(&self) -> Vec<&crate::core::Desc> {
+        vec![&self.desc]
+    }
+
+    fn collect(&self) -> Vec<crate::proto::MetricFamily> {
+        let mut m = MetricFamily::default();
+        m.set_name(self.desc.fq_name.clone());
+        m.set_help(self.desc.help.clone());
+        m.set_field_type(MetricType::GAUGE);
+        m.set_metric(RepeatedField::from_vec(vec![self.metric()]));
+        vec![m]
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::metrics::Collector;
+
+    #[test]
+    fn test_pulling_gauge() {
+        const VALUE: f64 = 10.0;
+
+        let gauge =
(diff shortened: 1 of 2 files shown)


JanBerktold:
@lucab hi! would you mind taking a look at this PR?

anacrolix:
We've been using this in production for a very long time. Please consider for merging!

JanBerktold:
Hey @lucab, would it be possible for you to take a look here? Thanks!

JanBerktold:
@lucab Hey there! We'd love to get this upstreamed if possible. Is there anything else I can do to help with this?

src/pulling_gauge.rs
```diff
@@ -0,0 +1,75 @@
+use std::{collections::HashMap, fmt, sync::Arc};
+
+use crate::{
+    core::Collector,
+    proto::{Gauge, Metric, MetricFamily, MetricType},
+};
+use protobuf::RepeatedField;
+
+/// A prometheus gauge that reads the value from a provided function on every
```
lucab, line 9: It would be better to split this docstring so that it has a short description. Like:
```rust
/// A [Gauge] that returns the value from a provided function on every collect run.
///
/// [... rest of text ...]
```

In the bottom section, you can even embed a small example. Also feel free to mention and link to the the Golang equivalent here.

src/pulling_gauge.rs
```diff
@@ -0,0 +1,75 @@
...
+            .finish()
+    }
+}
+
+impl PullingGauge {
+    /// Create a new PullingGauge.
+    pub fn new<S1: Into<String>, S2: Into<String>>(
+        fq_name: S1,
+        help: S2,
+        value: Box<dyn Fn() -> f64 + Send + Sync>,
```
lucab, line 32: A general question about the function here: does this assume that you have somewhere else some global mutable state? Or does this work with non-mutable captures too? (I'm mostly concerned about the lifetime of the underlying state)

src/pulling_gauge.rs
```diff
@@ -0,0 +1,75 @@
...
+            .field("desc", &self.desc)
+            .field("value", &"<opaque>")
+            .finish()
+    }
+}
+
+impl PullingGauge {
+    /// Create a new PullingGauge.
+    pub fn new<S1: Into<String>, S2: Into<String>>(
+        fq_name: S1,
```
lucab, line 30: What does `fq` stands for here? Let's maybe expand this name to make it obvious?

src/pulling_gauge.rs
```diff
@@ -0,0 +1,75 @@
...
+        value: Box<dyn Fn() -> f64 + Send + Sync>,
+    ) -> crate::Result<Self> {
+        Ok(PullingGauge {
+            value: Arc::new(value),
+            desc: crate::core::Desc::new(fq_name.into(), help.into(), Vec::new(), HashMap::new())?,
+        })
+    }
+
+    /// Create a new PullingGauge and register it with the global registry.
+    pub fn create_and_register<S1: Into<String>, S2: Into<String>>(
```
lucab, line 41: Unless you have a strong need for this, I'd rather drop this and leave the registration logic to library consumers.
The rationale is that the global default registry is an unfortunate thing, and I'd rather not push more users to that.

src/pulling_gauge.rs
```diff
@@ -0,0 +1,75 @@
...
+        metric
+    }
+}
+
+impl Collector for PullingGauge {
+    fn desc(&self) -> Vec<&crate::core::Desc> {
+        vec![&self.desc]
+    }
+
+    fn collect(&self) -> Vec<crate::proto::MetricFamily> {
```
lucab, line 67: I'd would be good to have a small end-to-end test, covering from `new()` till `collect()`.

JanBerktold: Yes! Added one below

src/pulling_gauge.rs
JanBerktold: Good callout, fixed.

src/pulling_gauge.rs
JanBerktold: It was supposed to stand for "fully qualified". However, I've removed this prefix since the other types don't have it either. It's just `name` now.

src/pulling_gauge.rs
JanBerktold: We generally are happy users of the global registry (we mostly write gRPC services using Rust for which it is convenient to have crates register standardized metrics _magically_) but I haven't actually found a place where we use this method in our internal code -- happy to remove it.

src/pulling_gauge.rs
JanBerktold: It does work with non-mutable captures too but typically we have some global state that we want to monitor. We typically use this to either
- fetch a system/runtime metric (e.g. std::thread::available_parallelism or tokio runtime metrics)
- report on the length of a queue by e.g. passing a `Arc<RwLock<>>` or an `ArcSwap`

JanBerktold:
Thanks for the comments! I think I've addressed all of them now -- Please take a look, @lucab.

anacrolix:
Thanks @lucab and @JanBerktold !

src/pulling_gauge.rs
```diff
@@ -0,0 +1,93 @@
...
+/// A [Gauge] that returns the value from a provided function on every collect run.
+///
+/// This metric is the equivalant of Go's
+/// <https://pkg.go.dev/github.com/prometheus/client_golang@v1.11.0/prometheus#GaugeFunc>
+///
+/// # Examples
+/// ```
+/// # use prometheus::{Registry, PullingGauge};
+/// let registry = Registry::new();
+/// let gauge = PullingGauge::new("available_parallelism", "The available parallelism, usually the numbers of logical cores.", Box::new(|| std::thread::available_parallelism().unwrap().get() as f64)).unwrap();
```
lucab, line 18: Thanks for adding a doc-example here, it's very useful. Perhaps it could be a bit more readable by splitting this line into multiple one and pulling out the polling function beforehand.

Unfortunately it looks like `std::thread::available_parallelism()` is more recent than our current MSRV. We could bump it here, or we could try to rework it. What about using `std::sync::Once::is_completed()` instead and showing it before and after completion?

lucab (commented):
Thanks for this, just one last minor nit on the doc example.

After that, it should be ready to merge, feel free to squash all of this into a single commit.

JanBerktold:
> Thanks for this, just one last minor nit on the doc example.
> 
> After that, it should be ready to merge, feel free to squash all of this into a single commit.

Should be good to go now! Changed the example & squashed all commits.

src/pulling_gauge.rs
lucab: Nice trick!

lucab (approved):
LGTM!

PR #411, opened 2021-09-01
Use vendored openssl

andylokandy:


andylokandy:
@marsishandsome @ekexium This patch help client-rust get rid of openssl dependency and therefore make client-py compatible on PyPI. I may provide a feature option for choosing vendering, but I think it should be ok so far.

ekexium:
I have no idea how this crate is used and whether it's a breaking change. An additional feature seems safer to me.

lucab:
Yes, I agree that changing the default features may not the best route to fix this.

To the best of my understanding, it may be the case that we don't even need to change anything here.
According to Cargo [feature unification rules](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification), it should be possible for a consumer  to directly declare a dependency on `reqwest` with the `native-tls-vendored` feature.
That should result in a single build of `reqwest` with the relevant TLS backend.
@andylokandy can you please try that route and see what's the outcome?

lucab (changes requested):
I'd suggest consumers to try exploring first doing this via cargo features unification, see https://github.com/tikv/rust-prometheus/pull/411#issuecomment-913455816.

andylokandy:
@lucab Good point! I've fixed the issue by that.

PR #416, opened 2021-10-04
Fix re-export of TEXT_FORMAT to not require protobuf

medcat:
A minor change, but previously the root-level re-export of `TEXT_FORMAT` would require `cfg(feature = "protobuf")`, despite the text export format not requiring protobufs.  Since this is the only location where `TEXT_FORMAT` is made public, it should be exported regardless of protobuf feature availability.


lucab:
Thanks for the PR, good catch!
Can you please amend your commit with a sign-off line? (See DCO check in CI for details)

PR #426, opened 2021-12-13
Fix broken doc links.

waywardmonkeys:
This fixes broken doc links by specifically telling it where to find the linked entity.


lucab (approved):
LGTM, thanks for the patch!

PR #430, opened 2021-12-28
feat(process): ProcessCollector use IntGauge to provide better perfor…

### LINKED ISSUE #429: feat(process): ProcessCollector use IntGauge to provide better performance
similar change like this PR in Tikv: https://github.com/tikv/tikv/pull/11308

pymongo:
similar change like this PR in Tikv: tikv/tikv#11308

close #429

Signed-off-by: wuaoxiang <wuaoxiang@stargraph.cn>

### DIFF
--- a/src/process_collector.rs
+++ b/src/process_collector.rs
@@ -4,13 +4,11 @@
 //!
 //! This module only supports **Linux** platform.
 
-use std::sync::Mutex;
-
 use lazy_static::lazy_static;
 
-use crate::counter::Counter;
+use crate::counter::IntCounter;
 use crate::desc::Desc;
-use crate::gauge::Gauge;
+use crate::gauge::IntGauge;
 use crate::metrics::{Collector, Opts};
 use crate::proto;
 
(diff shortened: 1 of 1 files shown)


lucab (approved):
LGTM, thanks for the patch!

src/process_collector.rs
```diff
@@ -27,13 +25,13 @@ const METRICS_NUMBER: usize = 7;
...
     pid: pid_t,
     descs: Vec<Desc>,
-    cpu_total: Mutex<Counter>,
-    open_fds: Gauge,
-    max_fds: Gauge,
-    vsize: Gauge,
-    rss: Gauge,
-    start_time: Gauge,
-    threads: Gauge,
+    cpu_total: IntCounter,
```
alin-at-dfinity, line 28: I totally get the performance argument, but could we please keep the `cpu_total` metric as `f64`? All the other metrics (file descriptor counts, memory usage, threads) are actual integers, but processes that don't use a lot of CPU will now have 1 second CPU usage spikes every so often instead of a more or less flat usage with actual spikes occurring when they actually do.

E.g. a process using 1% CPU on average, with a 10% spike every minute will now look as if it is using 0% CPU most of the time and spike every 2 minutes out of 3.

I really don't think the marginal performance improvement is worth it. Please keep using `Counter` for `cpu_total`.

src/process_collector.rs
```diff
@@ -90,7 +88,7 @@ impl ProcessCollector {
         .unwrap();
         descs.extend(rss.desc().into_iter().cloned());
 
-        let start_time = Gauge::with_opts(
+        let start_time = IntGauge::with_opts(
```
alin-at-dfinity, line 91: Actually, this is also not an integer. Although not as critical as CPU usage (because it is essentially a constant) the process start time has fractional seconds (and they may be useful).

PR #437, opened 2022-03-05
Derive default instead of obvious manual impl

jaseemabid:
The change is inconsequential, just something I noticed while reading the code.


breeswish (approved):
Thank you!

jaseemabid:
👋 @lucab

PR #446, opened 2022-06-24
process_collector: fix compilation on 32-bit targets

### LINKED ISSUE #442: cross-compilation broken for 32-bit builds
**Describe the bug**
when cross-compiling for 32-bit linux, the `libc::sysconf` call returns a 32-bit integer which is incompatible with the `i64` it wants to be stored in.

PR #430 made this change but it also removed the cast that made the previous version work on all platforms.

**To Reproduce**

Cross-compile using the `i686-unknown-linux-musl` target.

> Error:    --> /cargo/registry/src/github.com-1ecc6299db9ec823/prometheus-0.13.1/src/process_collector.rs:199:13
>        |
 >   197 |     static ref PAGESIZE: i64 = {
        |                          --- expected `i64` because of return type
    198 |         unsafe {
    199 |             libc::sysconf(libc::_SC_PAGESIZE)
        |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `i64`, found `i32`
        |
    help: you can convert an `i32` to an `i64`
        |
    199 |             libc::sysconf(libc::_SC_PAGESIZE).into()
        |                                              +++++++
    
>    For more information about this error, try `rustc --explain E0308`.
>    error: could not compile `prometheus` due to 2 previous errors

**Expected behavior**
There is no reason why it shouldn't be able to compile on 32-bit systems.

**System information**
* CPU architecture:
  * x86_64
* Distribution and kernel version: 
  * N/A
* SELinux on?:
  * N/A
* Any other system details we should know?:
  * Rust toolchain: ` stable-x86_64-unknown-linux-gnu unchanged - rustc 1.61.0 (fe5b13d68 2022-05-18)`

**Additional context**
Ran into this problem while building multi-arch docker images. It might be an idea to add cross-compilation to CI to catch problems like this in the future

jedisct1:
The crate didn't compile on 32-bit targets any more due to `libc::sysconf()` returning a 32-bit value on those targets.

Fixes #442

### DIFF
--- a/src/process_collector.rs
+++ b/src/process_collector.rs
@@ -190,14 +190,14 @@ lazy_static! {
     static ref CLK_TCK: i64 = {
         unsafe {
             libc::sysconf(libc::_SC_CLK_TCK)
-        }
+        }.into()
     };
 
     // getconf PAGESIZE
     static ref PAGESIZE: i64 = {
         unsafe {
             libc::sysconf(libc::_SC_PAGESIZE)
-        }
+        }.into()
     };
 }
 


lucab:
LGTM, thanks for the patch, can you please sign-off your commit?

jedisct1:
> LGTM, thanks for the patch, can you please sign-off your commit?

Done!

lucab (approved):

yinheli:
ping @lucab @overvenus, Could you please publish this to crates.io ?

jedisct1:
@yinheli meanwhile your can use the `prometheus-32bitfix` crate.

lucab:
Oh yes sorry, I wanted to tackle https://github.com/tikv/rust-prometheus/issues/451 too before releasing but never found proper time for that. Let's not delay this any longer and cut `0.13.2` directly, PR at https://github.com/tikv/rust-prometheus/pull/457.

PR #450, opened 2022-07-30
Feature: Replace lazy_static with once_cell

jaudiger:
Hi @lucab,

As suggested in https://github.com/tikv/rust-prometheus/issues/448, I did the migration from the lazy_static library to once_cell library.

I hope I didn't forget anything, let me know ! I also fixed a few warnings reported by clippy (I did it in another commit).

If there is anything else I could do, to help to merge this PR, ping me ! For example, I'm not sure, about having pushed the `proto_model.rs` file
Thanks in advance.


jaudiger:
@lucab , the CI seems to not pass due to the usage of Rust edition 2021 in one of the transient dependencies… I was going to open a such issue, using edition 2021 instead of 2018 seems to be required those days, since lot of projects already did the migration. What do you think about that ?

jaudiger:
Hi @lucab, do you have any feedback on this MR?

lucab:
Thanks for the PR! I did recently merge some CI/linting fixes that may spare you a lot of headaches here, you may try rebasing to current `master` and see if it helps. That way we likely don't need to switch to 2021 edition here. Also, CI requires signoffs on all commits here.
Rest of the changes looks good. I'd need to double-check whether this has implications on the semver/API side.

jaudiger:
@lucab thanks, I just rebased my MR on your latest commit. And I singed off my commits.

I also lower the restriction of the once_cell version to "^1.0" instead of "^1.13".

jaudiger:
Hi @lucab, any news regarding the implication of semver with those changes?

punkeel (approved):

PR #467, opened 2022-10-21
ci: add a job for minimum toolchain (MSRV)

lucab:
This adds a new CI job to verify building with a minimum supported Rust version (MSRV), currently set at 1.57.0.
This is done in preparation for a future edition-2021 switch.


lucab:
The `1.57.0` here is the result of our current dependency set (`cargo` verifies and enforces it when building).

BusyJay (approved):

PR #469, opened 2022-11-14
Add MaximumOverIntervalGauge

JanBerktold:
Depends on #405. 

This is another "fancy" gauge type from other internal codebase that I believe to be beneficial to others.

We have a few use cases where we want to monitor a value that moves around *a lot*. One example is measuring the number of concurrent requests for a gRPC server.
With a normal gauge, we're only able to inspect the value as of the time of the scraping (typically once every 10 seconds). This actually got us in production -- We had queue build ups due to downstream TCP re-connects that were manifesting themselves as increased latency for our customers but our metrics looked fine.

Enter the `MaximumOverIntervalGauge`: Instead of reporting the current value, we report the maximum of the value over the last (user configured) interval. This allows us to observe shifts that resolve themselves in less than the configured scrape interval.

### DIFF
--- a/src/maximum_over_interval_gauge.rs
+++ b/src/maximum_over_interval_gauge.rs
@@ -1,0 +1,212 @@
+use parking_lot::{RwLock, RwLockUpgradableReadGuard};
+use std::{
+    sync::{
+        atomic::{AtomicI64, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+#[cfg(test)]
+use mock_instant::Instant;
+
+#[cfg(not(test))]
+use std::time::Instant;
+
+use crate::{core::Collector, Error, PullingGauge};
+
+/// A prometheus gauge that exposes the maximum value of a gauge over an interval.
+///
+/// Used to expose instantaneous values that tend to move a lot within a small interval.
+///
+/// # Examples
+/// ```
+/// # use std::time::Duration;
+/// # use prometheus::{Registry, MaximumOverIntervalGauge};
+///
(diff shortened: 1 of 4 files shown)


src/maximum_over_interval_gauge.rs
```diff
@@ -0,0 +1,212 @@
...
+pub struct MaximumOverIntervalGauge {
+    // The current real-time value.
+    value: Arc<AtomicI64>,
+    // The maximum value in the current interval.
+    maximum_value: Arc<AtomicI64>,
+
+    // The length of a given interval.
+    interval_duration: Duration,
+    // The time at which the current interval will expose.
+    interval_expiry: Arc<RwLock<Instant>>,
```
JanBerktold, line 51: We actually use an `ArcSwap` for this in our internal codebase. I've switched this over to an `RwLock` here to avoid taking on the `ArcSwap` crate dependency.

src/maximum_over_interval_gauge.rs
```diff
@@ -0,0 +1,212 @@
...
+/// gauge.add(30);
+/// gauge.sub(10);
+///
+/// // For the next 30 seconds, the metric will be 30 as that was the maximum value.
+/// // Afterwards, it will drop to 10.
+/// ```
+#[derive(Clone, Debug)]
+pub struct MaximumOverIntervalGauge {
+    // The current real-time value.
+    value: Arc<AtomicI64>,
```
lucab, line 44: Here and everywhere else, gauges typically operate on `f64` values.

PR #471, opened 2022-12-12
Add `GenericCounter::set()` operation

tyranron:
## Synopsis

I've written the [`metrics-prometheus` crate](https://docs.rs/metrics-prometheus), trying to pave the way for uniting ecosystems of `prometheus` and `metrics` crates. While being somewhat complete, there are still minor issues which I cannot resolve (at least I didn't find a way) without making contributions to upstream.

One of them is that [`metrics::CounterFn` interface provides `absolute()` operation](https://docs.rs/metrics/0.20/metrics/trait.CounterFn.html#tymethod.absolute), while `prometheus::GenericCounter`'s API doesn't provide a way to [implement it atomically](https://docs.rs/metrics-prometheus/0.3.0/src/metrics_prometheus/metric.rs.html#53-63).

## Solution

This PR adds `GenericCounter::set()` operation, allowing to cover this use case.

## Checklist

- [x] Documentation added
- [x] Tests are added 
- [x] CHANGELOG entry is added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.13.4 (unreleased)
+
+- Add: `GenericCounter::set()` operation (#471)
+
 ## 0.13.3
 
 - Bug fix: Prevent ProcessCollector underflow with CPU time counter (#465)
--- a/src/counter.rs
+++ b/src/counter.rs
@@ -81,6 +81,17 @@ impl<P: Atomic> GenericCounter<P> {
         self.v.set(P::T::from_i64(0))
     }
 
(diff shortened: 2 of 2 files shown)


lucab:
Thanks for the PR, but please do note that in the larger Prometheus ecosystem by design Counters do not provide a `set()` method. They are meant to only increment, or reset back to 0 in very rare cases.
You can read more about that at https://www.robustperception.io/setting-a-prometheus-counter/ and https://www.robustperception.io/how-does-a-prometheus-counter-work/.

This was already discussed in https://github.com/tikv/rust-prometheus/pull/303 with enough consensus, as such I'm closing this PR.

PR #473, opened 2022-12-21
pulling_gauge: fix build with --no-default-features

tyranron:
## Synopsis

After merging #405 the `master` [fails](https://github.com/instrumentisto/metrics-prometheus-rs/actions/runs/3751488097/jobs/6372528212#step:4:80) to build [with `--no-default-features`](https://github.com/instrumentisto/metrics-prometheus-rs/pull/3/commits/f928edd4975a7d7b8acd24029aa532462e466bfb#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R23). The regression wasn't caught by CI as it doesn't check this scenario.

## Solution

- Fix the compilation issue by using `crate::from_vec!()` in `PullingGauge` implementation (as the other metric types do).
- Add commands to check `--no-default-features` scenario on CI.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -30,6 +30,8 @@ jobs:
         run: cargo build
       - name: cargo test
         run: cargo test
+      - name: cargo test (no default features)
+        run: cargo test --no-default-features
       - name: cargo test (extra features)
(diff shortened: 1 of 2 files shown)


lucab (approved):
LGTM, thanks for the fix and the CI improvement!

PR #479, opened 2023-03-06
 static-metric: fix typos in docstrings

eval-exec:
It should be `flashable` instead of `flash able`


lucab:
@eval-exec thanks for the PR, but the commit needs a `Signed-off-by` in order to pass CI.

lucab (approved):
Thanks for the patch, LGTM!

PR #497, opened 2023-09-05
Prevent `clippy::ignored_unit_patterns` in macro expansions

mweber15:
This is a pedantic lint added in 1.73.0. Because it occurs inside a macro expansion, the lint is triggered from user code. The justification given by the lint definition is:

> Matching with `()` explicitly instead of `_` outlines the fact that
> the pattern contains no data. Also it would detect a type change that
> `_` would ignore.

Removing the lint requires a trivial change. It does introduce the possibility for compilation errors in the event the return type of the function currently returning `()` changes, but that seems like more of a benefit than a drawback. In these cases, it seems unlikely that the return type in question will change in the future.

The user experience can be seen by linting the examples:

```
% cargo +nightly clippy --examples -- -A clippy::all -W clippy::ignored_unit_patterns -Z macro-backtrace
   Compiling prometheus v0.13.3 (/home/matt/src/rust-prometheus)
...
warning: matching over `()` is more explicit
   --> /home/matt/src/rust-prometheus/src/macros.rs:217:58
    |
214 |   macro_rules! register_counter {
    |   -----------------------------
    |   |
    |   in this expansion of `register_counter!` (#1)
    |   in this expansion of `register_counter!` (#2)
    |   in this expansion of `register_counter!` (#3)
...
217 |           $crate::register(Box::new(counter.clone())).map(|_| counter)
    |                                                            ^
...
221 |           register_counter!(@of_type Counter, $OPTS)
    |           ------------------------------------------ in this macro invocation (#3)
...
225 |           register_counter!(opts!($NAME, $HELP))
    |           -------------------------------------- in this macro invocation (#2)
    |
   ::: examples/example_push.rs:16:40
    |
16  |       static ref PUSH_COUNTER: Counter = register_counter!(
    |  ________________________________________-
17  | |         "example_push_total",
18  | |         "Total number of prometheus client pushed."
19  | |     )
    | |_____- in this macro invocation (#1)
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ignored_unit_patterns
    = note: requested on the command line with `-W clippy::ignored-unit-patterns`
...
```

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -214,7 +214,7 @@ fn test_histogram_opts_trailing_comma() {
 macro_rules! register_counter {
     (@of_type $TYPE:ident, $OPTS:expr) => {{
         let counter = $crate::$TYPE::with_opts($OPTS).unwrap();
-        $crate::register(Box::new(counter.clone())).map(|_| counter)
+        $crate::register(Box::new(counter.clone())).map(|()| counter)
     }};
 
     ($OPTS:expr $(,)?) => {{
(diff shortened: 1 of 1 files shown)


lucab (approved):
Thanks for the PR, LGTM!

PR #498, opened 2023-10-16
Add MaximumOverIntervalGauge

kushudai:
This is mostly a reimplementation of https://github.com/tikv/rust-prometheus/pull/469 updated to address the main comment.

The previous use case was fine using i64 as it was tracking queue sizes which were integral. However, we've noticed noise in our latency histograms. Since a `histogram_quantile` of 1 is merely an approximation (or an extrapolation, in case of not enough data points), we've seen "max" metrics skew heavily towards the the largest bucket.
We know these are not real because the server side timeouts are much smaller than the largest bucket but a bit bigger than the second largest bucket. 
A similar but separate problem is not having enough 9s for systems that serve hundreds of thousands of RPS - P99 does not accurately reflect tail latencies for these and adjusting the charts on a per use case basis is painful busywork. 

Instead of finely tuning buckets for different latency histogram metrics, we'd like to be able to report the maximum latency observed for a given time period (this is usually the scraping interval).
This allows us to put a cap on maximum latency seen on server side processing which then allows to accurately attribute network latency as seen by clients.


src/maximum_over_interval_gauge.rs
```diff
@@ -0,0 +1,236 @@
...
+            .expect("expected one MetricFamily to be returned");
+
+        let metric = metric_family
+            .get_metric()
+            .first()
+            .expect("expected one Metric to be returned");
+
+        assert_eq!(val, metric.get_gauge().get_value());
+    }
+}
```
kushudai, line 236:
```suggestion
}

```

kushudai:
Hi @lucab, I apologize for the nag but I was hoping you could take a look at this one :)

PR #506, opened 2023-12-07
Add type alias for unsigned integer gauge

DSharifi:
This PR exports a type of gauge for `AtomicU64`.  This is convenient when dealing with metrics that are of type `u64`.


DSharifi:
Hi @lucab! What do you think about exporting this type?

PR #511, opened 2024-01-17
cargo/examples: add required-features annotations

irevoire:
Hey,

I think we could use the [required features field](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field) in the cargo.toml to improve the general experience / error messages when trying to run an example which requires a specific feature


lucab (approved):
Thanks for the PR, LGTM!

PR #529, opened 2024-11-08
use `AsRef<str>` for slice of string labels

tabokie:


tabokie:
cc @overvenus @Connor1996

Connor1996 (approved):
LGTM

MrCroxx (approved):
LGTM

svanharmelen:
Any reason why this isn't merged yet (as it seems to be approved)? I struggle to create/set labels as they currently only take `&str`.

I get some values (which client connects) during a protocol handshake and tried to create a map with labels that I could store on the session and use everywhere that I need to update counters (something like`vec.with(&session.labels)`).

But that doesn't seem possible with the current options this crate gives for setting labels, but I think a PR like this would solve that and allow me to define the map as `HashMap<String, String>` instead of `HashMap<&str, &str>`. Right?

PR #532, opened 2024-12-12
Hashing improvements

hoxxep:
As part of investigating whether rust-prometheus would benefit from [rapidhash](https://github.com/hoxxep/rapidhash), I noticed a few improvements that could be made:

- The `MetricVecCore::children` hashmap uses keys that are hash values, so introducing `NoHashHasher` to avoid re-hashing this hash value improves performance.
- Allow `with` and `get_metric_with` to accept hashmaps with custom hashers.
- Introduce criterion's `black_box` in some hashmaps to reduce over-optimisation in microbenchmarks. Could be applied to further benchmarks.

Some of the `counter.rs` benchmarks are including hashmap initialisation and destruction. They are also using fixed values and small keys/values, which means benchmarks were unclear over whether FNV or rapidhash are beneficial (hash clashes, etc). Slightly longer key/value lengths are likely to benefit from rapidhash.


hoxxep:
@lucab sure! I've removed the nohash-hasher dependency and had a sparring session with the DCO signoff bot, but that should be done now 👍

src/nohash.rs
```diff
@@ -0,0 +1,23 @@
+use std::hash::{BuildHasherDefault, Hasher};
+
+/// Inspired by nohash-hasher, but we avoid the crate dependency because it's in public archive.
+#[derive(Copy, Clone, Debug, Default)]
+pub struct NoHashHasher(u64);
+
+pub type BuildNoHashHasher = BuildHasherDefault<NoHashHasher>;
```
lucab, line 7: I think these two can be more properly restricted to `pub(crate)`, right?

lucab (commented):
This generally looks fine, I only have a minor question.
It also needs a final `cargo fmt --all` in order to pass CI all checks.

src/nohash.rs
hoxxep: Yeah. They're technically never re-exported, but I've adjusted to `pub(crate)` per your request so that it's explicit to the reader.

lucab (approved):
LGTM

PR #537, opened 2025-03-03
Use AsRef<str> for owned label values

svanharmelen:
This PR is somewhat related to #529, but it's not trying to achieve/solve the same thing.

For a few of our applications we get some label values (e.g. client ID, client type, IP address) during a protocol handshake. Once the connection is setup each connection has a session object that remains valid during the lifetime of the session. So we were looking for a way to store the connection specifiek values in either a map or a slice so we could use them with `vec.with(&session.labels)` or `vec.with_label_values(&session.label_values)`.

But since the current implementation requires these maps or slices to only contain `&str` values, this currently isn't possible without some additional hacks or workarounds which makes the code less clean and less efficient.

Allowing values that implement `AsRef<str>` (so you can use `HashMap<&str, String>` maps or `Vec<String>` slices) solved this issue.

P.S. The second commit only contains a few LSP and Clippy related fixes to silence any warnings. Let me know if that is OK, if not I will drop that commit!


svanharmelen:
I noticed two checks failed, but it looks like its not related to my PR. Do you want me to try and fix it in my PR or will those be fixed in main instead?

lucab:
Thanks for the PR.
Yes CI is currently red, the fix for that is at https://github.com/tikv/rust-prometheus/pull/539 and pending review.

> P.S. The second commit only contains a few LSP and Clippy related fixes to silence any warnings. Let me know if that is OK, if not I will drop that commit!

That commit would better go into its own PR.

svanharmelen:
Thanks @lucab Removed the commit and created #540 for that 👍🏻

PR #541, opened 2025-03-12
cargo: upgrade to protobuf 3.7

### LINKED ISSUE #538: RUSTSEC-2024-0437: Update protobuf dependency
Now `cargo audit` has found vulnerability in `protobuf`.
Please update your `protobuf` dependency on actual v3.

```
$ cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 741 security advisories (from /home/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (528 crate dependencies)
Crate:     protobuf
Version:   2.28.0
Title:     Crash due to uncontrolled recursion in protobuf crate
Date:      2024-12-12
ID:        RUSTSEC-2024-0437
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0437
Solution:  No fixed upgrade is available!
Dependency tree:
protobuf 2.28.0
├── prometheus 0.13.4
│   ├── opentelemetry-prometheus 0.28.0
│   │   └── app 0.1.0
│   └── app 0.1.0
└── opentelemetry-prometheus 0.28.0

error: 1 vulnerability found!
```

sassman:
- fix #538 
- also closes #523
- keep some of the `get_` and `set_` methods around for compatibility reasons
- mark some of the `get_` methods as deprecated where there are successor methods
  - e.g. instead of using `Bucket::get_cumulative_count` use `Bucket::cumulative_count`
- keep API changes as minimal as positive
- introduce `MessageFieldExt` as an extension trait for `MessageField<Gauge>` and `MessageField<Counter>` to provide the `get_value()` method as used before, here without deprecation notice
- add convenience `From<>` implementations in the `proto_ext.rs` module to make the developer experience better when converting to the inner protobuf structs


sassman:
it appears that the Contributing guide does not exist.
How can I sign off the DCO?

image

sassman:
>  How can I sign off the DCO?

Figured it out

sassman:
put this PR to draft, because I figured that there are incompatibilities when the `protobuf` feature is not enabled.

sassman:
> put this PR to draft, because I figured that there are incompatibilities when the protobuf feature is not enabled.

This is now also fixed.

sassman:
If the changes regarding method deprecation are not wanted, I can also change it to comply fully to the non-protobuf models, and get rid of all the deprecation.

Please let me know what do you think about it.

lucab:
Thanks for the PR!
This is slightly large, I'll try to set aside some time next week to go through all of it.

> If the changes regarding method deprecation are not wanted, I can also change it to comply fully to the non-protobuf models, and get rid of all the deprecation.

To me, those looks nice to have. I don't expect many people to directly use those methods, but it's good to a buffer release with deprecation notices.

src/errors.rs
```diff
@@ -25,7 +25,7 @@ pub enum Error {
     /// An error containing a [`protobuf::error::ProtobufError`].
     #[cfg(feature = "protobuf")]
     #[error("Protobuf error: {0}")]
-    Protobuf(#[from] protobuf::error::ProtobufError),
+    Protobuf(#[from] protobuf::Error),
```
lucab, line 28: The docstring above needs to be updated in the same way too.

Cargo.toml
```diff
@@ -28,7 +28,7 @@ fnv = "^1.0"
 lazy_static = "^1.4"
 libc = { version = "^0.2", optional = true }
 parking_lot = "^0.12"
-protobuf = { version = "^2.0", optional = true }
+protobuf = { version = "^3.7", optional = true }
```
lucab, line 31: The security fix is actually in `3.7.2`, can we please directly put that as a lower limit in here?

src/lib.rs
```diff
@@ -129,23 +129,14 @@ This library supports four features:
...
 pub mod proto;
 
-#[cfg(not(feature = "protobuf"))]
-macro_rules! from_vec {
-    ($e: expr) => {
-        $e
-    };
-}
+#[cfg(feature = "protobuf")]
+mod convert;
```
lucab, line 137: Unless you have a strong feeling about keeping these two split, I'd move everything into a single `proto_ext` module.

lucab (commented):
This generally looks good, I've left a bunch of minor comments in review.

src/errors.rs
sassman: fixed with aae3a0df414fc6a7fe36a54c773e7d0a0c66ad12

Cargo.toml
sassman: good idea, done in aae3a0df414fc6a7fe36a54c773e7d0a0c66ad12

sassman:
Thanks for your quick feedback, I addressed all your changes.

lucab:
Thanks for the update.
CI is currently red due to some missing `#[cfg(feature = "protobuf")]` gating.
Other than that I think this is ready to go!

src/counter.rs
```diff
@@ -326,7 +326,10 @@ mod tests {
     use std::f64;
 
     use super::*;
-    use crate::metrics::{Collector, Opts};
+    use crate::{
+        metrics::{Collector, Opts},
+        proto_ext::MessageFieldExt,
```
lucab, line 331: This one seems to be missing a `#[cfg(feature = "protobuf")]`, I guess.

I think other test modules may also be broken in the same way under `--no-default-features`.

sassman: thanks, fixed now with bf6c3c1927c830640084ce7445ead32320240923

lucab (approved):
LGTM

PR #542, opened 2025-03-14
docs: enrich macro docs with more details/examples

rupert648:
Was doing some work with these macros and had a look and decided the docs could use a bit of love. Tried to add more examples and references to other items. Let me know what you think!


src/macros.rs
```diff
@@ -119,37 +260,118 @@ fn test_opts_trailing_comma() {
...
+///
+/// `HistogramOpts` defines the metadata and bucket configuration for histogram metrics.
+/// Histograms sample observations and count them in configurable buckets, which is essential
+/// for measuring distributions of values like request durations or response sizes.
+///
+/// # Syntax
+///
+/// This macro can be used in three ways:
+///
+/// ```
```
lucab, line 273: Here and below, it looks like some doc-tests are failing due to code mismatches (some missing `use` directives and some wrong API calls).
Can you run a `cargo test` and fix all the relevant issues?

lucab:
Thanks for the additional docs!
CI is currently red here, there are some doc-tests which need to be fixed.
It looks like the DCO bot expects a sign-off from `Rupert Carr <rupert@cloudflare.com>`, can you maybe update your commit?

PR #547, opened 2025-04-11
chore(histogram): Simplify code by using std lib method

Cardosaum:
Leverage existing standard library method to simplify implementation. No changes to public API or MSRV.

### DIFF
--- a/src/histogram.rs
+++ b/src/histogram.rs
@@ -1357,16 +1357,6 @@ mod tests {
     }
 
     #[test]
-    fn test_duration_to_seconds() {
-        let tbls = vec![(1000, 1.0), (1100, 1.1), (100_111, 100.111)];
-        for (millis, seconds) in tbls {
-            let d = Duration::from_millis(millis);
-            let v = d.as_secs_f64();
-            assert!((v - seconds).abs() < f64::EPSILON);
-        }
-    }
-
-    #[test]
     fn test_histogram_vec_with_label_values() {
         let vec = HistogramVec::new(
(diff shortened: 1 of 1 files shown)


src/histogram.rs
```diff
@@ -884,8 +884,7 @@ pub fn exponential_buckets(start: f64, factor: f64, count: usize) -> Result<Vec<
 /// `duration_to_seconds` converts Duration to seconds.
 #[inline]
 pub fn duration_to_seconds(d: Duration) -> f64 {
```
lucab, line 886: I think we can directly drop the `duration_to_seconds` function and replace it with `d.as_secs_f64()` at the calling sites.
Other than in tests, it is only used in one place, here:
https://github.com/tikv/rust-prometheus/blob/53d306635f232c4a2d9e7980da0262ec8a2c844e/src/histogram.rs#L543-L546

lucab (changes requested):

lucab:
@Cardosaum Thanks for the patch!
I'm happy to land something like this, but I think that it can be simplified even further while we are it.

src/histogram.rs
Cardosaum: Thank you for the feedback! 

Addressed the suggestion, please let me know if you think there's anything else that should be changed.

Cardosaum:
Thank you for the review @lucab! 
Addressed the changed requested, please let me know if there's anything else missing :)

lucab (approved):

PR #550, opened 2025-06-15
docs: Promote the usage of LazyLock instead of lazy_static

Expyron:
Since Rust 1.80, the recommended way to have lazily-initialized statics is to use [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html).

As this crate's MSRV is 1.81, this PR updates the documentation to encourage that new usage.

This is a change to examples and documentation only.
Once this PR is merged, I'll make another PR updating the internal uses of `lazy_static` within this crate.

### DIFF
--- a/examples/example_custom_registry.rs
+++ b/examples/example_custom_registry.rs
@@ -3,16 +3,14 @@
 //! This examples shows how to use multiple and custom registries,
 //! and how to perform registration across function boundaries.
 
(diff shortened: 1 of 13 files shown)


ThomasHabets:
https://github.com/tikv/rust-prometheus/pull/556 does this plus remove the dependency on `lazy_static` alltogether.

PR #553, opened 2025-07-30
Add a feature for push with reqwest rustls

silverfoxy:
I'm using this crate with "push" feature in an environment that requires boringssl instead of openssl. So my options for tls library are rustls or boring. With the current configuration, adding push feature pulls in reqwest with default features which relies on native-tls and openssl.

This PR is introducing a new feature (push-rustls) that allows push to work with rustls. I explicitly added the default features from reqwest that would be disabled by this change but do not affect the tls compatibility: https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L37C27-L37C61

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,8 @@ default = ["protobuf"]
 gen = ["protobuf-codegen"]
 nightly = ["libc"]
 process = ["libc", "procfs"]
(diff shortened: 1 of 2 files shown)


punkeel:
Hi @lucab, when you have a chance, could you help move this to completion? Thank you.

PR #564, opened 2026-03-28
Use floats for tracking cpu time

### LINKED ISSUE #560: process_cpu_seconds_total is incorrectly registered as an  IntCounter
https://github.com/tikv/rust-prometheus/blob/81514180fc47ed387d20140119b97c33495f85fe/src/process_collector.rs#L43C1-L53C1

```rust
        let cpu_total = IntCounter::with_opts(
            Opts::new(
                "process_cpu_seconds_total",
                "Total user and system CPU time spent in \
                 seconds.",
            )
            .namespace(namespace.clone()),
        )
        .unwrap();
        descs.extend(cpu_total.desc().into_iter().cloned());

```

This needs to be reported as a float, otherwise the rate queries are broken.  Consider that the promotheus server self-reports this metric in floats.

noncrab:
Fixes https://github.com/tikv/rust-prometheus/issues/560.

The one thing i'm not sure about is trading the well-defined wrap-around semantics for the risks of running out of bits in the mantissa. But then again, this uses an AtomicF64 under the hood, whose mantissa is 53 bits wide (can accurately hold an integer of up to ~9e+15).

Practically, by my (hand-wavey) reckoning, that'll mean we'll not be able to accurately represent times at ~1µs of precision after ~285 CPU millenia. I'm not sure how much of an issue that might be, or how best to record that concern.

I'd also be tempted to just use a gauge here as that'll avoid the need for calculating the delta since the last update (and the ensuing race condition) but that's bound to be a regression for someone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of CPU metrics through higher-precision calculations and enhanced handling of edge cases in metric computation.


noncrab:
@lucab Hi, is there any chance you (or someone else) could give this a look over and consider merging this please?

PR #567, opened 2026-08-06
Remove redundant casts

FranciscoTGouveia:
If I am not overlooking anything, I believe these casts are redundant and can thus be safely removed.

For reference, this change was found by an optimizer developed as a research project.

CC @nunoplopes

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal metric vector construction for counters, gauges, and histograms.
  * Preserved existing behavior and error handling.


FranciscoTGouveia:
I have tried to find the `CONTRIBUTING.md` file (both before sending the patch and after the bot's suggestion), but to no avail.
Am I missing something?

